Issue #32, opened 2022-10-07
Add syntax highlight to code blocks


sudobinbash:
Our current code blocks (converted from the previous doc) doesn't define the language for syntax highlight.

```markdown
    {
      "addr": "22",
      "proto": "tcp",
      "name": "ssh"
    }
```

resulting in a generic code block:

<img width="237" alt="image" src="https://user-images.githubusercontent.com/8906839/194675965-e29925d5-36fb-408d-86af-79b3ac3c81ce.png">

By adding the code block with language, we can enable syntax highlight

```markdown
 \```json
 {
     "addr": "22",
     "proto": "TCP",
     "name": "ssh"
 }
 \```
```

<img width="258" alt="image" src="https://user-images.githubusercontent.com/8906839/194676023-10b6012c-6656-49ff-a885-bb3b025d21c3.png">

nabin-nath:
@sudobinbash @russorat I want to contribute in improving the docs, can you assing this issue to me?

russorat:
i think is is working. closing for now

PR #121, opened 2022-12-07
added languages for highlighting syntax

nabin-nath:
closes https://github.com/ngrok/ngrok-docs/issues/32


sudobinbash:
hi @nabin-nath .  Unfortunately, your change is messing with line breaks in other pieces of code. It also doesn't solve for #32  entirely.

PR #127, opened 2022-12-12
Fix command for checking location of configuration file

nholden:
`verify` is not a valid command for `ngrok config`. This change updates the documentation to reference `ngrok config check`, which is a valid command that returns the location of the configuration file.


docs/ngrok-agent/config.md
```diff
@@ -392,4 +392,4 @@ Below is an example configuration file with all the options filled in.
           - region=us-east
           - team=infra
         addr: 8000
-        inspect: false
\ No newline at end of file
+        inspect: false
```
nholden: Looks like the GitHub editor removed a blank space from the end of this line. Let me know if that's an issue. 😅

sudobinbash (approved):
nice catch @nholden ! Thanks for contributing <3

PR #138, opened 2022-12-22
WordPress uses a capital P

pbrocks:
The capital P in WordPress is important enough that a [function](https://developer.wordpress.org/reference/functions/capital_p_dangit/) is created to correct this error in WordPress sites.


sudobinbash (approved):
Good Call @pbrocks . Approved 👍

PR #147, opened 2023-01-03
code snippets for docs/getting-started

avb-is-me:


docs/getting-started.md
```diff
@@ -107,21 +114,27 @@ You can test everything is working by running `ngrok -h` which should print the
...
 
 Start ngrok by running the following command.
 
-    ngrok http 8000
+```bash
+ngrok http 8000
+```
 
+```
 You should see something similar to the following console UI in your terminal.
```
sudobinbash, line 139: @avb-is-me . The code block should start after "You should see something similar to the following console UI in your terminal."

docs/guides/load-balancing-with-cloud-edges.md
```diff
@@ -22,9 +22,9 @@ Clicking the endpoint that was created when making this edge should open a new p
 Now let's start a tunnel to send traffic to your upstream service. You can click the "Start a Tunnel" button on this error page or look in the "Status" area of the HTTPS edge. Clicking this button will open a drawer with instructions on how to install and configure your agent.
 
 The command for starting the tunnel in the ngrok agent should look something like this:
-
-    ngrok tunnel 80 --region us --label edge=edghts_27RM7TKmP9FbDTFjT1uJuIyBgRk
-
+```
```
sudobinbash, line 25:
```suggestion
```bash
```

docs/guides/upgrade-v2-v3.md
```diff
@@ -7,13 +7,14 @@ The latest ngrok agent is packed with a ton of new capability but because it is
 ### Breaking Changes
 
 **Configuration file** - The older configuration file is not compatible with the latest ngrok agent without a few changes. We've automated the process to upgrade so all you need to do is run the new upgrade command.
-
-    ngrok config upgrade
+```
```
sudobinbash, line 10:
```suggestion
```bash
```

docs/guides/upgrade-v2-v3.md
```diff
@@ -7,13 +7,14 @@ The latest ngrok agent is packed with a ton of new capability but because it is
...
+ngrok config upgrade
+```
 
 **Unrecognized configuration parameters** - The previous ngrok agent silently ignored any config parameters it didn't understand. This means that you could have a configuration file with parameters that you thought were doing something, but they really weren't. This is also true for the ngrok agent API. The latest ngrok agent will explicitly error when it sees a configuration parameter that it doesn't recognize. It's a good idea to run a check after upgrading to find incompatible arguments.
-
-    ngrok config check
-
+```
```
sudobinbash, line 15:
```suggestion
```bash
```

sudobinbash (changes requested):
@avb-is-me . Just fix these and we should be good to go

docs/getting-started.md
avb-is-me: @sudobinbash done!

sudobinbash (approved):

PR #155, opened 2023-01-08
creates empty table headers

avb-is-me:
encapsulates current tables with empty headers with a div with a class of empty_headers_table and adds a css selector.  This way the only thing require for creating tables like this is to wrap it with this div.  Is the recommend approach for markdown.


sudobinbash:
Hi @avb-is-me 
This looks weird. I wonder if we can solve this using a React/MDX wrapper or something else.
I asked the Docusaurus community on Discord how they approach this... let's see what we get from there:
https://docusaurus.io/community/support

sudobinbash:
Hi @avb-is-me . This is what I got from the Docusaurus community:

> Correct - if you want headerless tables you need to find or write a remark plugin to support it.
> I think the one I use supports it.
> https://www.npmjs.com/package/remark-grid-tables

sudobinbash (changes requested):
t@avb-is-me Can you do this with a remark plugin instead? 

This is what I got from the docusaurus community:

> Correct - if you want headerless tables you need to find or write a remark plugin to support it.
> I think the one I use supports it.
> https://www.npmjs.com/package/remark-grid-tables


And this is how you expand docusaurus with remark plugins:

https://docusaurus.io/docs/markdown-features/plugins

sudobinbash:
Closing because we will go with remark tables.

PR #159, opened 2023-01-10
Alex/scroll to

alex-bezek:
Anchor links are not working in the site right now. If you turn on 3g network throttling, you can see it's actually rendering the main index page first before client side routing takes over and renders the correct page. This breaks the browser's native anchor link and hashtag functionality.

this is a hacky quick fix that makes it look like the anchor tags work without actually resolving the core issue. Its more meant as a stopgap until we can look into a better long-term solution


russorat (approved):
woah super hack-mode :)

alex-bezek:
> woah super hack-mode :)

yeah... not my favorite, but its quick and it works until we can find a better solution

ceschae (approved):
LGTM, thank you!!

PR #160, opened 2023-01-12
adds the remarkable table libary

avb-is-me:
Just adds the remarkable libary


sudobinbash:
recreate from the latest main

PR #185, opened 2023-02-02
Fix create credential request example

chetbox:
The request is required to be a POST but this seems to be missing from the example use of `curl`.


sudobinbash:
Ha! nice catch @chetbox !!! Thanks <333

sudobinbash (approved):

PR #189, opened 2023-02-08
fix: added hostedhooks and clerk

russorat:


sudobinbash (approved):

sudobinbash:
@faonbr 

As you publish new webhooks in our docs, would you mind updating the 
docs/cloud-edge/modules/webhook.md file accordingly?

Here's an example of change.

PR #190, opened 2023-02-09
Update config.md

ngrok-patrick:
I added an example of how this should be configured to the Large ngrok.yml example we reference


sudobinbash (approved):
LGTM!

PR #194, opened 2023-02-15
Change language around adding another user.

samcrichard:
Fixing per support ticket opened. This language was confusing.


russorat (approved):
lgtm

PR #203, opened 2023-02-21
feat(overview): convert to using ngrok-card

nijikokun:
Complete overhaul of the docs landing page to better present the options available to the user leveraging cards. 

Fully supports dark theme as well as mobile and includes some aria for links like changing color on hover and underlining links.

Other changes include:
- Introducing icons for top items
- Changing text to mention that you can use ngrok to secure application traffic
- Added Product / Reference sections with taglines for each product / section
- and more...

Screenshot 2023-02-21 at 1 45 29 PM

PR #206, opened 2023-02-22
Revert "feat(overview): convert to using ngrok-card"

nijikokun:
Reverts ngrok/ngrok-docs#203

Since we are serving content on S3 we can't use MDX

Docusaurus suggested not using S3


russorat (approved):
lgtm

PR #212, opened 2023-03-01
Adding new page for bot users in the ngrok platform section of the site

lee-tschetter-ngrok:
Adding new page for bot users in the ngrok platform section of the site and reordering the menu to fit


.gitignore
```diff
@@ -15,6 +15,7 @@
 .env.test.local
 .env.production.local
 .env
+yarn.lock
```
dev-cprice, line 18: you don't wanna ignore this

lee-tschetter-ngrok: Soooo I may have added that to my config locally and not fully realized that would try to push. 

But... Why not? Just for my own knowledge. A lock file is local to my machine when I run yarn start, right?

lee-tschetter-ngrok (commented):
Reverted .gitignore change.

.gitignore
dev-cprice: It describes how to recreate the packages tree in node_modules in a more reproduceable way. 

Some teams/people don't ever commit it or generate one (or npm/pnpm version), but what it ultimately means is you lose the ability to pin dependency versions (and dependencies of your dependencies).

This can suck when you don't wanna update to the latest minor/patch version of dependencies that either: shipped breakages or didn't respect semver (both happen way more often than they should)

PR #219, opened 2023-03-06
fix: adding trailing slash

russorat:
Adds a trailing slash to our docs urls in the sitemap and when navigating to avoid a 302 redirect.


nijikokun (commented):
LGTM!

PR #227, opened 2023-03-09
Simplifies Integration Contributor Process

avb-is-me:
On build we run a node js script to look at all the integration folders -> creates a json called integrationItems.json -> concats that json with an existing json copy of prior integrations(this can actually be removed but thought it would be nice to keep a copy) -> uses that integration Items json in the navbar in sidebar.js


nijikokun:
Might want to hold off on this PR as we will be changing the structure of integrations soon, this does look like a step in the right direction though!

nijikokun:
@avb-is-me just pushed out a PR which simplifies the contributor process for integrations, no more need to edit `sidebar.js`.  Just need to ensure that any frontmatter values in `webhooks.md` or `index.mdx` values exist and are correct.

See #239

avb-is-me:
@nijikokun, nice!  I'll close

PR #232, opened 2023-03-14
fix: point to ssh reverse tunnel page

nijikokun:
Instead of pointing to the ngrok used for ssh page, point users to the reverse tunnel docs page.


russorat (approved):
LGTM

PR #233, opened 2023-03-14
fix: updating to use v2 commands

russorat:
somehow the changes from a previous commit were lost. bringing them back over.


nijikokun (approved):
LGTM!

PR #249, opened 2023-03-21
feat: Add example of using docker compose

### LINKED ISSUE #201: add docker compose example to the docs
we should have a page with an example docker-compose file that a user could use.

russorat:
closes #201


docs/using-ngrok-with/docker.md
```diff
@@ -14,3 +14,24 @@ Note: for MacOS or Windows users, the `--net=host` option will not work. You wil
 ```bash
 docker run -it -e NGROK_AUTHTOKEN=xyz ngrok/ngrok:latest http host.docker.internal:80
 ```
+
+# Using ngrok with Docker Compose
+------------
+
+If you're more comfortable using Docker Compose, you can use the following as a starting point. Copy the contents below into a new file named `docker-compose.yaml`, then you can  and run `docker compose up` in the directory. This Docker compose file assumes that you have an `ngrok.yml` file in the same directory with at least one tunnel defined. Check out the [ngrok agent config file documentation](/docs/ngrok-agent/config/) for help creating a configuration file with a tunnel definition. If you want to use the same configuration file as your local ngrok agent, you can view the location of the default config file using `ngrok config check`.
```
russorat, line 21:
```suggestion
If you're more comfortable using Docker Compose, you can use the following as a starting point. Copy the contents below into a new file named `docker-compose.yaml`, then run `docker compose up` in that directory. This Docker compose file assumes that you have an `ngrok.yml` file in the same directory with at least one tunnel defined. Check out the [ngrok agent config file documentation](/docs/ngrok-agent/config/) for help creating a configuration file with a tunnel definition. If you want to use the same configuration file as your local ngrok agent, you can view the location of the default config file using `ngrok config check`.
```

nijikokun (approved):
Looks good as a starting point!

PR #261, opened 2023-03-25

ceschae:
This PR is from the docs gen pipeline from internal ngrok workings. It includes several changes to docs that were implemented in features but not included here yet. There are 2 action items from reviewers:

- [ ] A decision on whether or not to write over the manually edited API header links
- [ ] Look at the `yarn.lock` file and let me know what to do about that.

2 PRs will follow up this one:

1. A PR to add "This is a generated file, do not edit" at the top of all generated files
2. A PR to switch from the manual error list to the generated one


docs/api/resources/abuse-reports.mdx
```diff
@@ -75,7 +75,7 @@ Returns a 200 response on success
 | `urls` | List&lt;string&gt; | a list of URLs containing suspected abusive content |
 | `metadata` | string | arbitrary user-defined data about this abuse report. Optional, max 4096 bytes. |
 | `status` | string | Indicates whether ngrok has processed the abuse report. one of `PENDING`, `PROCESSED`, or `PARTIALLY_PROCESSED` |
-| `hostnames` | [AbuseReportHostname](#abusereporthostname-fields-1) | an array of hostname statuses related to the report |
+| `hostnames` | [AbuseReportHostname](#abusereporthostname-fields) | an array of hostname statuses related to the report |
```
ceschae, line 78: TODO: a lot of these `-1` links get trimmed with the auto generation. The good news is that with the edits to the gen process, there will be _zero_ broken links in the API docs. The bad news is they always go to the first table. **Would you rather leave these manually edited, or wait until this can be permanently fixed?** This is a non-trivial problem to fix, to know where to add the `-1`, '-2', etc

docs/errors/_err_ngrok_list.md
ceschae, line 1: TODO: I've built and served this locally and we can replace the `_manual` list with this one. I'd recommend that be a separate PR though, so I'll tag you in that one once it's out.

yarn.lock
ceschae, line 1: TODO: I have no clue why this got edited in the gen process. Maybe you can provide some context for me about these changes, if this will need a manual ignore?

russorat (approved):
i mean, there's a lot of stuff here. feels like we merge this and look for bugs to address afterwards.

russorat:
specifically:

A decision on whether or not to write over the manually edited API header links
 - i think you should overwrite

I don't know what to do about yarn.lock, but i'd say overwrite and we will fix it if it's broken. it looks like most of it is package updates.

PR #268, opened 2023-03-29

ceschae:
- adds line to generated files that it is generated
- picks up a new error
- picks up an API field's description


ceschae:
weird that it closed but whatever :shrug: i'll re-open once i finalize the changes on the gen size

PR #273, opened 2023-04-04
k8s-ingress-controller-tutorial

sudobinbash:


docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+    ```
+
+1. Install the ngrok Ingress Controller in your cluster, replacing `[AUTHTOKEN]` and `[API_KEY]` with your Authtoken and API key from above:
+
+    ```bash
+    export NGROK_AUTHTOKEN=[AUTHTOKEN]
+    export NGROK_API_KEY=[API_KEY]
+    helm install ngrok-ingress-controller ngrok/kubernetes-ingress-controller \
+    --set credentials.apiKey=$NGROK_API_KEY \
+    --set credentials.authtoken=$NGROK_AUTHTOKEN
```
alex-bezek, line 57: nit: usually the next lines are indented by a tab to make it more clear these are the same line wrapping

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+    kubectl apply -f ngrok-manifest.yaml
```
alex-bezek, line 58: this seems like it should be removed since the file is created in the next step

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+    ```yaml showLineNumbers
+    apiVersion: v1
+    kind: Service
+    metadata:
+    name: game-2048
+    spec:
+    ports:
+        - name: http
+        port: 80
+        targetPort: 80
```
alex-bezek, line 77: I think you have some tab/spacing issues 
image

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+    :::tip Notes:
+    - Lines 1-32: Create the 2048 app service and deployment
+    - Lines 34-51: Create the ngrok Ingress Controller. Line 42 in specific determines the ingress URL for public requests.
+    :::
+
+    ```yaml showLineNumbers
+    apiVersion: v1
+    kind: Service
+    metadata:
+    name: game-2048
```
alex-bezek, line 72: one thing that might help people avoid namespace problems like you had would be if every resource specifies the namespace, otherwise it uses whatever namespace they have set in their context. 

Everything not having a namespace should make them all end up in the same namepace asuming someone doesn't change namespace halfway through, but if they do and the manifets have namespaces set, they'll take precedence.

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+    ---
+    # ngrok Ingress Controller Configuration
+    apiVersion: networking.k8s.io/v1
+    kind: Ingress
+    metadata:
+      name: game-2048-ingress
+    spec:
+      ingressClassName: ngrok
+      rules:
+        - host: my-awesome-k8s-cluster.ngrok.app
```
alex-bezek, line 110: might make it clear that this needs to be templated. some guides will also include a sed command to find/replace things

`sed -i "s/__PILLAR__LOCAL__DNS__/$localdns/g; s/__PILLAR__DNS__DOMAIN__/$domain/g; s/__PILLAR__DNS__SERVER__/$kubedns/g" nodelocaldns.yaml`

https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+## Step 3: Add edge security to your app
+
+The ngrok Ingress Controler for Kubernetes provides custom resource definitions (CRDs) for additional edge features available in ngrok. In this example, we're expanding the Ingress Controller with Google OAuth to allow access only from users with the email domains `@acme.com` or `@ngrok.com` and to apply a circuit breaker to your app at 80%. These features are enforced at the ngrok edge, ensuring only authorized users can access your app.
+
+:::tip Notes:
+This example is very similar to the previous version, with the following changes:
+- Lines 40-42: Associates your Ingress Controller with the configuration `oauth-and-circuit-breaking` via annotation
+- Lines 56-72: Sets the edge configuration as a custom CRD (NgrokModuleSet).
+- Lines 62-67: Sets the circuit breaker module over 50% threshold.
+- Lines 68-72: Sets the OAuth module to allow access only for users with the email domains `@acme.com` or `@ngrok.com`.
```
alex-bezek, line 145: hmmm, these would probably be easier to follow and keep up to date if we could just show a diff between the files with some callout, but i don't know of a great tool for that tbh. The service and deployment aren't changed, maybe cutting down that noise would help here. 

Something like this maybe shows more clearly the diff but makes it harder to copy paste the right code

image

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+        app: game-2048
+    spec:
+      containers:
+        - name: backend
+          image: alexwhen/docker-2048
+          ports:
+            - name: http
+              containerPort: 80
+---
+# Instead of running the k8s ingress on port 80
```
alex-bezek, line 182: extra comment?

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+      - acme.com
+      - ngrok.com
+---
+```
+
+1. Save your file and reapply the manifest file:
+    
+    ```bash
+    kubectl delete -f ngrok-manifest.yaml
+    kubectl apply -f ngrok-manifest.yaml
```
alex-bezek, line 228: No need to do a delete here, just an apply should skip the resources with no changes and update the ones that did saying they are configured

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
+---
+title: Kubernetes
+---
+
+# Using ngrok with Kubernetes 
+------------
+
+## Introduction
+
+The [ngrok Ingress Controller for Kubernetes](https://github.com/ngrok/kubernetes-ingress-controller) is our official open-source controller for adding public and secure ingress traffic to your k8s clusters. If you’ve used ngrok in the past, you can think of the ngrok Ingress Controller as ngrok packaged an idiomatic k8s resource — available as helm chart, configurable via manifest (using the `kind: Ingress` construct), and compatible with k8s best practices.
```
alex-bezek, line 10: Nits:
`to your k8s clusters`: technically you are giving access to the workloads/services in the cluster, not the cluster itself if that distinction is important

`you can think of the ngrok Ingress Controller as ngrok packaged an idiomatic k8s resource`: controllers in general are kubernetes components you can run to extend the functionality of k8s. Ingress controllers are specifically meant to manage `kind: Ingress` resources to convert the standard declarative Ingress Spec model into the controller _implementation specific_ method of providing ingress to a service.

docs/using-ngrok-with/k8s.md
```diff
@@ -0,0 +1,246 @@
...
+1. An [ngrok account](https://ngrok.com/signup) with a Pro subscription. The subscription is required in order to configure the ngrok Ingress Controller with a custom domain and an Edge Configuration.
+:::
+
+## Get started with the ngrok Ingress Controller for Kubernetes
+
+## Step 1: Get your ngrok authtoken and API key
+
+:::tip
+The ngrok Ingress Controller requires:
+- The authtoken to launch connections from ngrok's points of presence to your k8s cluster.
```
alex-bezek, line 25: do we normally talk about points of presence or just "the edge servers"?
I'm also not sure which way we normally say its connecting, but i envision it as connecting from the ingress controller in your k8s cluster, to the edge servers to allow traffic to back down the other way

docs/using-ngrok-with/k8s.md
sudobinbash: nice catch!

docs/using-ngrok-with/k8s.md
sudobinbash: I <3 YAML

docs/using-ngrok-with/k8s.md
sudobinbash: Yeah. Docusaurus doesn't provide diff ootb. 

The closest I can get is a line highlighter like in go.
https://ngrok-docs-git-frederico-ngrok-ingress-controller-ngrok-dev.vercel.app/docs/using-ngrok-with/go/

docs/using-ngrok-with/k8s.md
sudobinbash: removing! good call.

docs/using-ngrok-with/k8s.md
sudobinbash: nice!

docs/using-ngrok-with/k8s.md
sudobinbash: changed the copy. Feel free to make suggestions.

docs/using-ngrok-with/k8s.md
sudobinbash: changed the copy. Feel free to make suggestions.

docs/using-ngrok-with/k8s.md
sudobinbash: added namespace for all resources. Also brought back create a namespace to the helm install command

docs/using-ngrok-with/k8s.md
sudobinbash: Oh... I'll need some help understanding this (these pesky marketers 😄  )

docs/using-ngrok-with/k8s.md
sudobinbash: The new version uses highlight. Unfortunately we cannot have just the diff without compromising the copy/pasta.

docs/using-ngrok-with/k8s.md
alex-bezek: sorry, thats an overly complex example of using sed tbh 😅 
we can worry about it later. I assume in the long run we'll want to have a consistent way of people doing things like substituting values and we can work on that consistency over time

docs/using-ngrok-with/k8s.md
alex-bezek: oooh a built in way with docusaurus one day would be cool. also just a future thought

alex-bezek (approved):
This is great, much nicer than the markdown readme files we started with!

PR #290, opened 2023-04-08
Updated Okta webhook doc to new layout

faonbr:
Okta webhooks don't support HMAC., as per https://developer.okta.com/docs/concepts/event-hooks/#security


docs/integrations/okta/webhooks.md
```diff
@@ -2,83 +2,133 @@
...
+
+This will get the project installed locally.
+
+Now you can launch the app by running the following command: 
+
+```bash
+npm run startOkta
+```
+
+The app runs by default on port 3000.
```
nijikokun, line 49:
```suggestion
The app runs by default on port `3000`. 
```

russorat (approved):
lgtm

PR #295, opened 2023-04-12
ketch: reduce block padding on action buttons in the consent banner

cody-dot-js:
### Before

image

### After

image

image

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -470,7 +470,8 @@ a .ngrok--card svg {
  * operation: make the banner smaller
 */
 #lanyard_root [class^="content"] {
-  padding-block: 12px;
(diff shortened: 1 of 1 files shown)


src/css/custom.css
```diff
@@ -470,7 +470,8 @@ a .ngrok--card svg {
  * operation: make the banner smaller
 */
 #lanyard_root [class^="content"] {
-  padding-block: 12px;
+  padding-top: 12px;
+  padding-bottom: 12px;
```
cody-dot-js, line 474: padding-block has less browser support than I would like for a global/public facing site 😞

KristopherPaulsen (approved):
ε=ε=ε=┌( >_<)┘

PR #303, opened 2023-04-19
feat: add prettier

### LINKED ISSUE #297: Add prettier fmt
https://prettier.io/docs/en/install.html

should prefer these defaults:

```js
// prettier.config.js
module.exports = {
	arrowParens: "always",
	semi: true,
	singleQuote: false,
	tabWidth: 2,
	trailingComma: "es5",
	useTabs: true,
};
```

russorat:
closes #297


package.json
cody-dot-js, line 1: only one thing missing: ci scripts to format

let's add the following scripts under `"scripts"`:

```json
"cmd:prettier": "pnpm exec prettier --cache --config ./prettier.config.js --ignore-path ./.prettierignore",
"fmt:check": "pnpm run cmd:prettier --check .",
"fmt": "pnpm run cmd:prettier --write .",
```

cody-dot-js (approved):

package.json
cody-dot-js: oh, we should probably also add `pnpm run fmt:check` to the ci checklist

likely just add it here: https://github.com/ngrok/ngrok-docs/blob/5599194fcdc6397f81a06d62455b2a09989cd5cc/.github/workflows/pr.yml#L7-L26

cody-dot-js: feel free to punt on both of these ☝️ 

we can add them in a follow up

PR #328, opened 2023-05-08
feat: add terminology doc

nijikokun:
Introduces a terminology document to the Cloud Edges section with frequently discussed concepts. This is very basic at the moment and is likely to expand as documentation is updated / changed.


russorat:
@nijikokun this might not be needed now

PR #330, opened 2023-05-12
use direnv/nix/flakes to setup env

nikolay-ngrok:
Use direnv, flake and nix to deliver nodejs and pnpm when you get into the docs directory.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,7 +24,7 @@
       {
         devShell = pkgs.mkShell {
           shellHook = ''
-            export PATH="$PATH:node_modules/.bin"
+            export PATH="$PATH:$(git rev-parse --show-toplevel)/node_modules/.bin"
           '';
           buildInputs = with pkgs; [
             node-toolchain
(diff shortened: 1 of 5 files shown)


hjkatz (approved):

nijikokun:
@nikolay-ngrok @hjkatz can you better explain what flake & direnv is used for here?

flake.nix
```diff
@@ -0,0 +1,34 @@
...
+        };
+        node-toolchain = with pkgs; [
+          nodejs
+          nodePackages.pnpm
+        ];
+      in
+      {
+        devShell = pkgs.mkShell {
+          shellHook = ''
+            export PATH="$PATH:node_modules/.bin"
```
jrobsonchase, line 27: :grimacing: 

relative paths in the `PATH` variable are a security hazard. Can we `git rev-parse --show-toplevel` to pin it to the repo root?

jrobsonchase (approved):
I'm a little surprised that the nested `node-toolchain` list works, but if you're getting the packages installed in your shell, I guess it's fine :shrug:

PR #331, opened 2023-05-13
fix: adding static images for the docker desktop extension

russorat:
just some images so that i can embed them into the manifest for the docker desktop extension.


nijikokun:
Is the readme also going to be embedded?

russorat:
i just added that so future us knows what these are. but also, i'm not sure this is going to work as i can't find these images on the preview. maybe there's a better way

sudobinbash (approved):
LGTM

PR #338, opened 2023-05-31
add clearbit tracking

sudobinbash:


nijikokun (approved):
LGTM

For posterity sake -- looks like it [must be loaded synchronously](https://help.clearbit.com/hc/en-us/articles/360063559573-Install-the-Clearbit-Tag-) and when referrers are set it won't show the source unless it's served via the correct domain.

PR #345, opened 2023-06-22
Pull API documentation out into a top-level section.

inconshreveable:
This PR seeks to take an initial step in improving our documentation by refactoring the API documentation.

- Visibility: API documentation was hidden underneath the ngrok Platform section making it hard to find.
- Unification: API documentation and API reference documentation were separated into two top-level sections.
- Defragmentation: API documentation was unnecessarily separated into many separated pages without a clear reason which makes it harder to read and search the documentation.
- Succinctness/Clarity: As I made changes to fix the above, I ended up trimming the API documentation to remove extraneous words, fix typos, adjust header sizes,  and do other small quality-of-life fixes.

As part of this change, I needed an API reference index page to link to so I created one that talks a little bit about how we generate the API reference documentation.

I'm intending to make similar improvements for the remainder of the documentation.

It's possible this may break a few links to some lesser-referenced pages in our documentation. I'm not sure yet how to solve this problem and could use guidance here.

This also deletes an `apis` directory in the root of this repository which from what I can see was completely unused and confusing.


docs/api/index.md
```diff
@@ -1,39 +1,129 @@
 ---
-sidebar_position: 1
-title: Introduction
+sidebar\_position: 1
```
nijikokun, line 2: Is the `\` here intentional or accidental?

docs/api/index.md
```diff
@@ -1,39 +1,129 @@
...
 
-The ngrok HTTP API is available at `https://api.ngrok.com`. It provides programmatic access to all of ngrok's resources.
+## Overview
 
-The API is REST-ish. It follows most of the conventions of a REST API but breaks some when the REST model does not fit well. The API listens only on port 443 to help avoid any accidental unencrypted requests.
+The ngrok API provides programmatic access to all of ngrok's resources.
 
-If you are looking to programmatically start and stop tunnels, instead consult the documentation of the [ngrok agent API](/ngrok-agent/api).
+The API is REST-ish. It follows most of the conventions of a REST API but
+diverges slightly when the REST model does not fit well.
```
nijikokun, line 14: Nice shift in framing 👍

docs/api/reference.md
```diff
@@ -0,0 +1,26 @@
+---
+sidebar\_position: 1
```
nijikokun, line 2: Same here, accidental?

docs/api/index.md
inconshreveable: ty!

docs/api/index.md
inconshreveable: my editor tells me that this is the right thing to do when editing a markdown doc. i think it's confused because it's the yaml front matter but in a markdown document you need to escape an underscore if you don't want to use it for _emphasis_

inconshreveable:
@nijikokun can i get an approval? do you know how to redirect for the old pages?

nijikokun:
@inconshreveable I was going to suggest doing that but it looks like the API urls would stay at the same location.

https://ngrok.com/docs/api/

inconshreveable:
they don't actually as far as i can tell, but i did find a way to do redirects! updating momentarily

docs/api/index.md
nijikokun: Ah. Let's remove them in the front-matter to be consistent with the rest of the codebase

nijikokun:
Error from deploy:

> [ERROR] Error: Invalid sidebar file at "sidebars.js".
> --
> 15:42:07.175 | These sidebar document ids do not exist:
> 15:42:07.175 | - events/event-sources
> 15:42:07.175 | - events/event-sources

nijikokun (approved):
LGTM 🚀

PR #346, opened 2023-06-23
fix: documentation for oauth headers

jonstacks:
Small fix. The documentation has the returned header listed as `ngrok-auth-email`, but it is actually `ngrok-auth-user-email`.


inconshreveable (approved):
ty!

PR #348, opened 2023-06-23
Fix for next page links

wdawson:
I noticed a few pages had broken `next` links where it would be pointing to the page itself. It looks like this is because of how docusaurus treats sidebar categories that act as their own link rather than just a category. I went through all the categories to make sure I got all of those. 

With this change, it will be trickier to maintain these links, but we'll preserve the nice UX for users clicking on the category name in the sidebar (1 click vs 2).


docs/integrations/index.mdx
```diff
@@ -1,9 +1,10 @@
 ---
+pagination_next: integrations/aftership/index
```
nijikokun, line 2: I'm wondering if we disable for this, since this page is fully generated.

package.json
```diff
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@docusaurus/module-type-aliases": "^2.4.0",
+    "@docusaurus/plugin-content-docs": "^2.4.1",
```
nijikokun, line 39: What's this being used for?

wdawson: it's in [sidebar.js](https://github.com/ngrok/ngrok-docs/blob/main/sidebars.js#L14)

docs/integrations/index.mdx
wdawson: Yeah, this was the biggest issue one. I can make it `null`

nijikokun (approved):
LGTM failing test happens when integrations gets touched so assuming that's what is happening here.

wdawson:
failing test looks unrelated:
```
Status: Downloaded newer image for algolia/docsearch-scraper:latest
docker.io/algolia/docsearch-scraper:latest
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/root/src/index.py", line 119, in <module>
    run_config(environ['CONFIG'])
  File "/root/src/index.py", line 45, in run_config
    config.query_rules
  File "/root/src/algolia_helper.py", line 13, in __init__
    self.algolia_client = SearchClient.create(app_id, api_key)
  File "/root/.local/share/virtualenvs/root-BuDEOXnJ/lib/python3.6/site-packages/algoliasearch/search_client.py", line 51, in create
    config = SearchConfig(app_id, api_key)
  File "/root/.local/share/virtualenvs/root-BuDEOXnJ/lib/python3.6/site-packages/algoliasearch/configs.py", line 63, in __init__
    super(SearchConfig, self).__init__(app_id, api_key)
  File "/root/.local/share/virtualenvs/root-BuDEOXnJ/lib/python3.6/site-packages/algoliasearch/configs.py", line 22, in __init__
    assert app_id, "app_id cannot be empty."
AssertionError: app_id cannot be empty.
Error: Process completed with exit code 1.
```

PR #360, opened 2023-07-10
Rework Integrations Docs Organization

inconshreveable:
This PR is a proposal for consideration. If we want to adopt it, there's a bunch of additional work to do before this can merge.  This is just a sketch.

It refactors our Integrations documentation to do three things:

1. Organize our integrations by *how* they integrate instead of strictly by vendor. image
2. Merge about half of `Using ngrok with . . .` into `Integrations`. I didn't move the remainder of the items in `Using ngrok with . . .` because many of them are not exactly integrations, per se, although you could make an argument for some of them. I'm not sure where those will find a home eventually but I think that's a problem for later.
3. Remove the 'nesting' of integration providers. Previously, for each vendor you would need to click into the vendor and then again to get to a guide. Now you only need to do that if an integration vendor has more than one guide. I didn't finish moving them all, but I did enough to give us an idea of what the change would feel like:
image

If we want to move forward, I'll create a complete TODO list of what needs to be done to complete the transition.


russorat:
closing in favor of https://github.com/ngrok/ngrok-docs/pull/1105

PR #364, opened 2023-07-11
Update docker.md - examples

ignorabilis:
Hi, I was wondering how the different options are provided to the ngrok agent and wasted a bit of time until I finally found out that the convention is basically the same as the agent one - so I bumped the docs to make it more obvious.


inconshreveable (approved):
love this, ty so much!

PR #375, opened 2023-07-19
split out the errors reference into its own page, clean up copy

inconshreveable:
this updates the errors section of the documentation to have the same structure as Events and API with two separate pages: Overview and Reference. consistency in how each documentation section is organized makes it easier for developers to find what they're looking for

i improved the copy a little bit as well while i was in here


mkarnowski (approved):
adding approval

PR #379, opened 2023-07-24
Use a timestamped tmp index name

krwenholz:
The timestamped index resolves an issue in which GitHub CI will sometimes run the index job twice. This causes the whole indexing job to fail under certain race conditions and our site to have _no_ indexing.

Also correct a few formatting issues.

Issue seen [here](https://github.com/algolia/docsearch-scraper/issues/426) per a recommendation from Algolia Support.


russorat (approved):
amazing! lgtm

PR #383, opened 2023-07-27
chore(config.md): use env=prod instead of region

nijikokun:
The example mentions region under labels which causes confusion for new users thinking that setting region in the label will affect the agent region.


krwenholz (approved):
Can you get the one on line 341 as well, please?

nijikokun:
> Can you get the one on line 341 as well, please?

Done!

PR #385, opened 2023-07-27
fix: broken links and add security pdf

nijikokun:
- Fixes broken links pointing at `https://ngrok.com/static`
- Adds missing PDF under static folder


russorat (approved):
lgtm

PR #388, opened 2023-07-28

natasha-jarus:
Mostly intended to update api docs for event destinations, adding datadog. 

There's a lot here to review. What's worth reviewing:

- `api/resources/*.mdx`
- `errors/*.md`

what's ignorable:
- `api/resources/toc/`
- `api/resources/examples/`
- `events/`


nijikokun:
I think we may need to update the generation to [better handle headers](https://ngrok-docs-git-ngrokrock-dev-gen-changes-ngrok-dev.vercel.app/docs/api/resources/agent-ingresses/). Also the auto-generated message causes the syntax highlighting to break which is unfortunate.

natasha-jarus:
further investigation: we use mdx v1 which only supports xml-style comments, so these will have to do for now. it is unfortunate that github's syntax highlighting doesn't understand that, but it's not a dealbreaker imo for generated code.

nijikokun (approved):

PR #399, opened 2023-08-14
Refactor header module documentation

inconshreveable:
This changeset refactors the documentation for the request headers and response headers modules. Header template variables are factored out into a partial and included into each page.

Including the headers in each page unfortunately doesn't allow you to see subnavigation for the different sections of the variables. This is a bug in docusaurus. It may make sense to eventually move them back out into their own page if we can eventually find a home for them. This is better for the time being.

The section on automatic headers belongs in a more generic section about the http reverse proxy behaviors of ngrok's edge because they happen regardless of whether this module is enabled or not. This section will eventually be moved when a more appropriate place for that type of documentation exists.


docs/cloud-edge/modules/_http_header_templates.mdx
```diff
@@ -1,62 +1,46 @@
...
-
-### Circuit Breaker Template Variables {#circuit-breaker}
-
-Note: These values are only available when using the Circuit Breaker module.
+These variables are only available when using the Circuit Breaker module.
 
 |                                |                                                                                                                                                                                                         |
 | ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `${.circuit_breaker.decision}` | Whether the HTTP request was sent to the upstream service. 'allow' if the breaker was closed, 'block' if the breaker was open, 'allow_while_open' if the request was allowed while the breaker is open. |
+| `${.circuit_breaker.decision}` | Whether the HTTP request was sent to the upstream service. `allow` if the breaker was closed, `block` if the breaker was open, 'allow_while_open' if the request was allowed while the breaker is open. |
```
russorat, line 32:
```suggestion
| `${.circuit_breaker.decision}` | Whether the HTTP request was sent to the upstream service. `allow` if the breaker was closed, `block` if the breaker was open, `allow_while_open` if the request was allowed while the breaker is open. |
```

docs/cloud-edge/modules/_http_header_templates.mdx
```diff
@@ -1,62 +1,46 @@
...
 
-### Compression Template Variables {#compression}
+#### Compression Variables {#compression}
 
-Note: These values are only available when using the Compression module.
+These variables are only populated when using the Compression module.
 
 |                               |                                                                                                            |
 | ----------------------------- | ---------------------------------------------------------------------------------------------------------- |
 | `${.compression.algorithm}`   | The compression algorithm used to encode responses from the endpoint. Either 'gzip', 'deflate', or 'none'. |
```
russorat, line 40:
```suggestion
| `${.compression.algorithm}`   | The compression algorithm used to encode responses from the endpoint. Either `gzip`, `deflate`, or `none`. |
```

docs/cloud-edge/modules/_http_header_templates.mdx
```diff
@@ -71,9 +55,7 @@ Note: These values are only available when using the Compression module.
 | `${.conn.server_port}` | The port that the connection for this request came in on.                                       |
 | `${.conn.start_ts}`    | The timestamp when the TCP connection to the ngrok edge is established.                         |
 
-### Geo IP Template Variables {#ngrok-geo}
-
-This data is provided by [MaxMind GeoIP](https://www.maxmind.com/en/home).
```
russorat, line 76: any specific reason to remove this line? i know we've been asked about where this data comes from in the past.

docs/cloud-edge/modules/_http_header_templates.mdx
```diff
@@ -97,26 +79,35 @@ This data is provided by [MaxMind GeoIP](https://www.maxmind.com/en/home).
...
 
-### IP Policy Template Variables {#ip-policy}
+#### IP Restrictions Variables {#ip-restrictions}
 
-Note: These values are only available when using the IP Policy module.
+These variables are only populated when using the IP Restrictions module.
 
 |                               |                                                                                               |
 | ----------------------------- | --------------------------------------------------------------------------------------------- |
 | `${.ip_policy.decision}`      | 'allow' if IP Policy module permitted the request to the upstream service, 'block' otherwise. |
```
russorat, line 88:
```suggestion
| `${.ip_policy.decision}`      | `allow` if IP Policy module permitted the request to the upstream service, `block` otherwise. |
```

docs/cloud-edge/modules/_http_header_templates.mdx
```diff
@@ -97,26 +79,35 @@ This data is provided by [MaxMind GeoIP](https://www.maxmind.com/en/home).
...
 
-|                          |                                                  |
-| ------------------------ | ------------------------------------------------ |
-| `${.ngrok.client_ip}`    | This is the original client IP of the request.   |
-| `${.ngrok.request_id}`   | This is the unique request Id generated by ngrok |
-| `${.ngrok.saml.subject}` | The SAML subject of the the authenticated user.  |
+|                        |                                                  |
+| ---------------------- | ------------------------------------------------ |
+| `${.ngrok.client_ip}`  | This is the original client IP of the request.   |
+| `${.ngrok.request_id}` | This is the unique request Id generated by ngrok |
```
russorat, line 96:
```suggestion
| `${.ngrok.request_id}` | This is the unique request ID generated by ngrok |
```

docs/cloud-edge/modules/_http_header_templates.mdx
```diff
@@ -126,21 +117,39 @@ Note: These values are only available when using the OAuth module.
...
-| `${.tls.client_cert.serial_number}` | The serial number of the client's leaf TLS certificate in the Mutual TLS handshake.       |
-| `${.tls.client_cert.subject.cn}`    | The subject common name of the client's leaf TLS certificate in the Mutual TLS handshake. |
-| `${.tls.client_cert.subject}`       | The subject of the client's leaf TLS certificate in the Mutual TLS handshake              |
-| `${.tls.version}`                   | The version of the TLS protocol used between the client and the ngrok edge.               |
+|                           |                                                                                                        |
+| ------------------------- | ------------------------------------------------------------------------------------------------------ |
+| `${.oauth.app_client_id}` | The is the ID of the OAuth application used to handle this request.                                    |
+| `${.oauth.decision}`      | 'allow' if the OpenID Connect module permitted the request to the upstream service, 'block' otherwise. |
```
russorat, line 128:
```suggestion
| `${.oauth.decision}`      | `allow` if the OpenID Connect module permitted the request to the upstream service, `block` otherwise. |
```

docs/cloud-edge/modules/request-headers.mdx
```diff
@@ -0,0 +1,349 @@
...
+
+There is a bug which currently causes the above behavior to not be correct.
+Only the last header to be used when specifying multiple headers. This behavior
+will be fixed to match what is documented above.
+
+:::
+
+If you remove a header that has multiple values, all values will be removed.
+
+### Replacing Header Values
```
russorat, line 201: this also doesn't function properly due to the same reason above. i'm leaning on just adding these sections in when the bug is fixed.

docs/cloud-edge/modules/response-headers.mdx
```diff
@@ -0,0 +1,344 @@
...
+
+```yaml
+tunnels:
+  example:
+    proto: http
+    addr: 80
+    response_headers:
+      add:
+        [
+          "content-security-policy: self",
```
russorat, line 38: obviously both are valid, but i think we use the hyphen syntax for arrays in yaml in other areas of the docs:

```
    response_headers:
      add:
        - "content-security-policy: self"
        - "dial-duration: ${.backend.dial_duration}"
      remove:
        - internal-trace-id
```

russorat (approved):
see comments. otherwise, LGTM

docs/cloud-edge/modules/request-headers.mdx
inconshreveable: replacing doesn't work or removing multiple values doesn't work?

inconshreveable: as far as i can tell this works

docs/cloud-edge/modules/response-headers.mdx
inconshreveable: fixed

docs/cloud-edge/modules/request-headers.mdx
```diff
@@ -0,0 +1,349 @@
...
+```
+
+The client will receive the following:
+
+```http
+GET / HTTP/1.1
+host: example.ngrok.app
+foo: original-value
+foo: new-value
+```
```
russorat, line 225: this example, this would not contain both values. it would only contain the `foo: new-value` header due to the bug above.

docs/cloud-edge/modules/_http_header_templates.mdx
```diff
@@ -97,26 +81,35 @@ This data is provided by [MaxMind GeoIP](https://www.maxmind.com/en/home).
...
+These variables are only populated when using the IP Restrictions module.
 
 |                               |                                                                                               |
 | ----------------------------- | --------------------------------------------------------------------------------------------- |
-| `${.ip_policy.decision}`      | 'allow' if IP Policy module permitted the request to the upstream service, 'block' otherwise. |
+| `${.ip_policy.decision}`      | `allow` if IP Policy module permitted the request to the upstream service, `block` otherwise. |
 | `${.ip_policy.matching_rule}` | The rule that triggered an IP Policy match on the endpoint.                                   |
 
-### ngrok Template Variables {#ngrok-variables}
+#### ngrok Variables {#ngrok-variables}
```
rkolavo, line 93: It seems like these variable sections are in alphabetical order except the ngrok variables have jumped ahead of Mutual TLS.

docs/cloud-edge/modules/response-headers.mdx
```diff
@@ -0,0 +1,344 @@
...
+```http
+HTTP/2 200
+foo: bar
+foo: baz
+```
+
+:::note
+
+There is a bug which currently causes the above behavior to not be correct.
+Only the last header to be used when specifying multiple headers. This behavior
```
rkolavo, line 197:
```suggestion
Only the last header will be used when specifying multiple headers. This behavior
```

docs/cloud-edge/modules/response-headers.mdx
```diff
@@ -0,0 +1,344 @@
...
+
+```http
+HTTP/2
+foo: new-value
+```
+
+### Case Sensitivity
+
+When adding headers, ngrok normalizes all header keys to a lower case
+reprsentation per the http/2 RFC. See [RFC
```
rkolavo, line 245:
```suggestion
representation per the http/2 RFC. See [RFC
```

docs/cloud-edge/modules/request-headers.mdx
```diff
@@ -0,0 +1,349 @@
...
+GET / HTTP/1.1
+host: example.ngrok.app
+foo: bar
+foo: baz
+```
+
+:::note
+
+There is a bug which currently causes the above behavior to not be correct.
+Only the last header to be used when specifying multiple headers. This behavior
```
rkolavo, line 194:
```suggestion
Only the last header will be used when specifying multiple headers. This behavior
```

docs/cloud-edge/modules/request-headers.mdx
```diff
@@ -0,0 +1,349 @@
...
+```http
+GET / HTTP/1.1
+host: example.ngrok.app
+foo: new-value
+```
+
+### Case Sensitivity
+
+When adding headers, ngrok normalizes all header keys to a lower case
+reprsentation per the http/2 RFC. See [RFC
```
rkolavo, line 245:
```suggestion
representation per the http/2 RFC. See [RFC
```

docs/cloud-edge/modules/request-headers.mdx
```diff
@@ -0,0 +1,349 @@
...
+
+If you remove a header that has multiple values, all values will be removed.
+
+### Replacing Header Values
+
+If you add a header that is already present in the HTTP request, it will add
+another header. For example, if you run:
+
+```bash
+ngrok http --request-header-add "foo: new-value"
```
russorat, line 207:
```suggestion
ngrok http 80 --request-header-add "foo: new-value"
```

docs/cloud-edge/modules/_http_header_templates.mdx
inconshreveable: fixed, ty.

@russorat unrelated low-priority item: it feels like these `.ngrok` scoped variables should be deprecated in favor of variables that already exist in other namespaces

`.ngrok.client_ip` -> `.conn.client_ip`
`.ngrok.request_id` -> `.http.request_id`

`.http.request_id` i think should become `.http.request.ngrok_id` as well to match the scoping of all of the other http request variables which are all nested under `.http.request`. right now `.request_id` is an outlier.

doing this migration should be easy. we should just be able to add these new paths and keep the old ones around but remove their public documentation

PR #401, opened 2023-08-15
Reorganize existing modules to match new format.

inconshreveable:
There are no meaningful changes in this changeset, it just moves around sections and text in the docs of existing modules to match the new format used by the docs for request/response headers and the oauth modules.

It also adds syntax highlighting to existing examples + normalizes how we display variables that need to be added by the user in example code.


russorat (approved):
LGTM

PR #432, opened 2023-10-04
update documentation for region field description to mark it as "deprecated"

KristopherPaulsen:
## Why

Global network is upon us. We don't want to be confusing or misleading customers by having outdated documentation. Lets update it!

## Validation

Checked all the places where the old text was present, and ran locally

Eg: Example below

image


ForgoneReality (approved):
your delivery of rubber stamp has arrived

PR #433, opened 2023-10-05
Correctly map custom font weights

aaronshekey:
This PR modifies our font declarations to properly map to our font weights. Currently, we don't map Euclid Semibold to the `600` weight. This causes browsers to render a faux bold which affects readability. This PR fixes that by spec'ing those weights to a single `EuclidSquare` font name, and changing font files per weight.

In the overview, you'll see lots of faux bold fonts get changed to a real semibold. You'll notice the top menu also got a little thicker, since it was intended to be the "medium" font weight. Now it's actually rendered as such.

## Before
Screenshot 2023-10-05 at 9 43 56 AM

## After
Screenshot 2023-10-05 at 9 45 15 AM

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -5,37 +5,35 @@
  */
 
 @font-face {
-	font-family: "EuclidSquare";
+	font-family: EuclidSquare;
+	font-style: normal;
+	font-weight: normal;
 	font-display: swap;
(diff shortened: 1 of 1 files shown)


russorat (approved):
yesssss! lgtm

PR #439, opened 2023-10-13
change configuration event name to audit events

mkarnowski:
change configuration event name to audit events to incorporate additional events sources


russorat (approved):
LGTM

PR #440, opened 2023-10-14
Markdown text colors

aaronshekey:
We were only applying our text color to the paragraph element. This spec's that color on the entire markdown body, and then applies a darker color to headers only. This should solve the issue of lists and tables have darker text than paragraphs.


russorat (approved):
lgtm!

PR #448, opened 2023-10-17
Add upstream option for agent terminate_at/terminate-at setting in agent docs

ctindel:
The agent docs for the CLI and YAML config files were missing the "upstream" option for the terminate_at / terminate-at settings.


russorat (approved):
lgtm!

PR #450, opened 2023-10-18

cody-dot-js:


russorat:
i think the path has now switched from `docs/events` to `docs/obs`. can we update that in the buildkite job as well?

PR #459, opened 2023-10-24
chore(k8s): make registering a domain an explicit step

nijikokun:
We've gotten a lot of feedback that it is not clear you must register the domain through the dashboard. This PR makes it so that is required step before you get to setting the domain field and explains to the user they will need it later.


russorat (approved):
LGTM

PR #475, opened 2023-11-06
feat: switching over to new docsearch

russorat:
we are moving over to the algolia search where they crawl and index our site from their end. should improve docs search like 1000000%

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,17 +40,3 @@ jobs:
           aws s3 sync ./build s3://docs-s3.dev-ngrok.com/docs --acl public-read
           aws s3 sync ./build s3://docs-s3.stage-ngrok.com/docs --acl public-read
           aws s3 sync ./build s3://docs-s3.ngrok.com/docs --acl public-read
-  search-index-prod:
-    needs: deploy
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - name: Run algolia/docsearch-scraper
-        run: |
-          docker pull algolia/docsearch-scraper
-          docker run \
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -39,18 +39,4 @@ jobs:
...
-      - name: Run algolia/docsearch-scraper
-        run: |
-          docker pull algolia/docsearch-scraper
-          docker run \
-            -e ALGOLIA_APP_ID=${{ secrets.DEV_ALGOLIA_APPLICATION_ID }} \
-            -e API_KEY=${{ secrets.DEV_ALGOLIA_API_KEY }} \
-            -e INDEX_NAME_TMP="prod_ngrok_tmp_$(date +%s)"  \
-            -e "CONFIG=$(cat ./algolia_prod.config.json | jq -r tostring)" \
-            algolia/docsearch-scraper
+          aws s3 sync ./build s3://docs-s3.ngrok.com/docs --acl public-read
```
krwenholz, line 42: This shouldn't show as a newline. I think the tabbing is off?

krwenholz (changes requested):
I think indentation is off for the workflows. Otherwise looks correct!

.github/workflows/deploy.yml
russorat: that's strange, the tabbing looks all good on my side. not sure how to fix

krwenholz: Weird. If it looks okay locally and no mixed tabs/spaces, that's okay!

krwenholz (approved):
We chatted. Mine and GitHub's eyes must just be bad.

PR #478, opened 2023-11-09
Top navigation refinements

aaronshekey:
This PR unifies our navigation with the current marketing site while making visual refinements that improve the responsiveness. Prior to this PR, if the window was of a certain size, the navigation would overlap the logo. This also refines the log in and sign up buttons at the desktop and mobile sizes.

image

Current navigation:
image


russorat (approved):
thank you!

PR #479, opened 2023-11-10
Left navigation refinements

aaronshekey:
This pull request explores some design refinements in our left navigation. It adds section names and addresses some hierarchy issues. This PR also serves as a prototype to evaluate against static mockups.

image

image

Pop around on the Vercel build to check it out :v:


inconshreveable:
love this. ty so much aaron! couple thoughts:

- in light mode, contrast between nav section header and nav section items is challenging (for my eyes at least). this is especially true in the endpoints section where the small caps treatment isn't a differentiator. the contrast is improved in dark mode.
- thoughts on a different treatment for an active 'section'? ("Network Edge" in your screenshot) the bolded but still de-emphasized font weight feels a little odd to my eyes
- if it doesn't look weird, it would be nice to pull FAQ out into its own unlabeled section at the bottom

russorat (approved):
lgtm!

Issue #481, opened 2023-11-14
labels only available for not HTTPv2Tunnel


HugoGresse:
According to the doc, we should be able to set the labels on tunnels: https://ngrok.com/docs/agent/config/#labeled-tunnel-configuration 

So a config.yml like this: 
```
authtoken: xxxxxxxxxxxxx
region: eu
version: "2"
tunnels:
    api:
        labels:
         - edge=xxxxxxxxxxxxxxx
        addr: 3000
        proto: http
        domain: https://xxxxxxx.ngrok-free.app
```

should be doable, 

But it result in error : `field labels not found in type config.HTTPv2Tunnel` 
using `ngrok config check --config config.yml`

If this is expected, can the doc be clarified? 

`ngrok v3.4.0`

inconshreveable:
remove both `proto: http` and `domain: ...`

you'll see in the linked documentation that neither of those properties are supported in labeled tunnel configuration. instead, when you used labeled tunnels, the protocol and domain are configured by the matching edge

HugoGresse:
Thanks for the clarification, can I make a small PR to reflect this on the doc?

inconshreveable:
please!

PR #483, opened 2023-11-14
Update config.mdx with tip about labels and domains

HugoGresse:
The `domain` and `protocol` don't go along with the `labels`, so the doc better reflect that

See discussion in https://github.com/ngrok/ngrok-docs/issues/481


russorat (approved):
lgtm

PR #486, opened 2023-11-14
Fixing minor typo in Authtoken section

salilsub:
The original text said "When you provision a new authtoken, the full token is only once.", which is an incomplete sentence. Updated the sentence to indicate that it is displayed once.


russorat (approved):
lgtm

PR #488, opened 2023-11-15
Update JS SDK links

salilsub:
We updated the JS SDK from ngrok-nodejs to ngrok-javascript. Updating the links since docs link did not forward and was broken.

### DIFF
--- a/docs/agent-sdks/index.mdx
+++ b/docs/agent-sdks/index.mdx
@@ -40,13 +40,13 @@ ngrok's edge just as if you opened a socket to listen on a port.
 
 ## Supported Languages
 
-| Language   | Docs                                                       | Quickstart                                      | Repository                                                             | Status |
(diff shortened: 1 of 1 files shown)


russorat:
Could you also add a redirect in redirect.js from the old page to the new page? This helps anybody who is deep linking

bobzilladev:
> Could you also add a redirect in redirect.js from the old page to the new page? This helps anybody who is deep linking.
> this file: https://github.com/ngrok/ngrok-docs/blob/main/static/scripts/fix-redirect.js

This is a link out to github pages, so I don't think that redirect applies. It's sad github isn't redirecting this for us.

bobzilladev (approved):

salilsub:
The repo link does redirect, just not the Github pages/gitub.io link

PR #496, opened 2023-11-18
tcp tunnels: region is no longer required with the advent of cross-region tcp tunnels

natasha-jarus:
Now folks don't need to care about regions for TCP tunnels!

See https://github.com/ngrok-private/ngrok/issues/23493


tammybailey (approved):
I trust your judgement on this - my only concern would be that we've removed everything! You might want to tag russ for review as niji is OOO for a while.

russorat (approved):
Love it!

PR #507, opened 2023-11-23
rename nodejs -> javascript

inconshreveable:
- naming is now consistent
- all documentation links work again (they are all currently broken because the github repository was renamed)


russorat (approved):
LGTM!

PR #508, opened 2023-11-23
document API CORS behaviors

inconshreveable:
our API supports CORS but we've never documented how. document the API's CORS behaviors


russorat (approved):
LGTM

PR #510, opened 2023-12-01
changelog for agent release 3.5.0

bobzilladev:


salilsub:
Do we state anywhere what the valid options are for app-protocol? If it's in the `--help` command that's sufficient.

bobzilladev:
> Do we state anywhere what the valid options are for app-protocol? If it's in the `--help` command that's sufficient.

Not currently: `--app-protocol string              Set application protocol`

It doesn't do local validation since we want that to be an error from the server side so we don't have to constantly update it. But since this feature isn't totally launched yet this is a bit vague.

salilsub:
I agree that we should not validate locally but I do think we should state somewhere what the accepted values are. More so that a user knows what the right values are rather than validating whether what they enter is correct.

bobzilladev:
Added a note about the `--app-protocol` value.

salilsub (approved):

PR #511, opened 2023-12-03
Update typo for the usage of the CLI `ngrok http` command 

kevinlutzer:
The additional option for selecting what schemes to forward on is `--scheme` not `--schemes`

Screenshot 2023-12-03 at 1 18 56 AM


inconshreveable (approved):
nice catch kevin, ty for catching my mistake!

PR #518, opened 2023-12-05
add ~/.ssh/id_ed25519.pub to ssh-reverse-tunnel-agent

bobzilladev:
Mention `id_ed25519.pub` since that is the ssh default now.


russorat (approved):
lgtm!

PR #526, opened 2023-12-07
Automated generation of docs from buildkite

wdawson:
Primarily driven by wanting to add new edge modules to docs, but this picks up a bunch more stuff too, which are all expected:

- Most of the examples have dates and IDs changed. A few new examples for jwt and user agent filter modules
- Various new and updated errors got picked up from this generation
- Various events have been changed as well

## Unexpected change

There are a lot of these added to various files:
```md
---
```

I suspect this is from the templating code. Should I go back and remove those from the template? Or do we want the divider?


wdawson:
From slack convo, removing the dividers was intentional. Standby for force push.

Megalonia (approved):

PR #527, opened 2023-12-07
batch of formatting fixes (missed in docusaurus v3 upgrade)

cody-dot-js:
indented code blocks don't render correctly now (since all `.md` and `.mdx` are parsed w/ mdx)

image

https://mdxjs.com/docs/what-is-mdx/#markdown


aaronshekey (approved):
Thanks for the additional fixes here! Nice to see a language applied to more and more of these.

PR #530, opened 2023-12-08
Wils/sidebar edge fix

wdawson:
- Replaces edge module API reference docs with files we are actually generating.
- I suspect the old stuff is a holdover from a while ago
- Moves all references.


russorat (approved):
lgtm! thanks

PR #536, opened 2023-12-14
fix: event tables

russorat:
this is a temp fix until we fix the generators for these tables.


salilsub:
The top row/header is blank on a lot of the tables, should that be the case?

aaronshekey (approved):

PR #538, opened 2023-12-16
Not showing integration page if only one integration type exists.

faonbr:


S3Prototype:
Closed because 2 years old

PR #539, opened 2023-12-18
Update mentions of sdk's on using-ngrok-with pages

bobzilladev:
Move the SDK's up to top position, remove outdated alpha/beta notes, add links for java.

### DIFF
--- a/docs/using-ngrok-with/django.md
+++ b/docs/using-ngrok-with/django.md
@@ -8,13 +8,9 @@ title: Django
 This article assumes you have Python, PIP, and Django already installed.
 :::
 
-### Agent Management
+### Python SDK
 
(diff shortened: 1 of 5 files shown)


salilsub:
Do we want to even keep the "Agent Management" sections? It feels like they are competing with the SDKs and we probably want to route users to our stuff rather than libraries managed and maintained by 3rd parties.

bobzilladev:
> Do we want to even keep the "Agent Management" sections?

True, I left them around when our first-party libraries were still in development, but I'd be happy to remove em now.

salilsub:
The Java SDK is done with development, right? I know we are still working on blogs and examples, but the actual SDK development is done?

salilsub (approved):

PR #467, opened 2023-10-31
Add integration guide: Ingress to applications managed by Rancher in Kubernetes

joelhans:
In a similar vein to the work I previously did on #429, this PR adds a guide for setting up a local/demo Rancher installation, deploying a new RKE2 cluster, and then installing the ngrok Ingress Controller plus a sample app on that cluster.

PR #429, opened 2023-10-03
Add guide: Ingress to services in a vcluster on Kubernetes

joelhans:
Hello! This PR adds a guide for integrating the Kubernetes Ingress Controller with a vcluster running locally on Minikube. We can think of is as a companion guide to the [vcluster edition](https://loft.sh/blog/using-the-ngrok-ingress-controller-to-create-preview-environments-with-vcluster/) those folks created a few months back.

PR #543, opened 2023-12-21
Add integration guide: Ingress to a microservices-based application connected via Linkerd's service mesh

joelhans:
Part of the same series of work as #467 & #429—this PR adds a guide on how to integrate the ngrok Kubernetes Ingress Controller with Linkerd and an example `Emojivoto` app.


stmcallister (approved):
This looks great! 🎉

PR #556, opened 2024-01-11
Update flake.lock so node version is compatible with package.json

jonstacks:
When using nix, I noticed warnings about unsupported engine

### Before

image

https://github.com/ngrok/ngrok-docs/blob/fcc89a2d2f6bbdbeb14c44984b8f0175b64e7d14/package.json#L6-L8

I ran `nix flake update` to update the version of nixpkgs we are using which now puts us on node 20.10.0.

### After flake.nix update

image

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1681202837,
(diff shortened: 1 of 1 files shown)


nikolay-ngrok (approved):
👍

PR #563, opened 2024-01-18
use updated method of signing repos

browner12:
`apt-key` and `/etc/apt/trusted.gpg.d` are deprecated due to security concerns.  the new recommended method is to directly reference the key in the repository declaration.

one possible issue users could run into is if the `/etc/apt/keyrings` directory is not already created, but it's an easy enough fix.

some references for the deprecation and changes:

- https://wiki.debian.org/DebianRepository/UseThirdParty
- https://www.linuxuprising.com/2021/01/apt-key-is-deprecated-how-to-add.html
- https://askubuntu.com/questions/1286545/what-commands-exactly-should-replace-the-deprecated-apt-key?newreg=c7d51f5afac945fea79db7a504d68e76

### DIFF
--- a/docs/getting-started/index.mdx
+++ b/docs/getting-started/index.mdx
@@ -26,8 +26,8 @@ brew install ngrok/ngrok/ngrok
 
 ```bash
 curl -s https://ngrok-agent.s3.amazonaws.com/ngrok.asc | \
-  sudo tee /etc/apt/trusted.gpg.d/ngrok.asc >/dev/null && \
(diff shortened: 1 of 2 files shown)


andrewryno (approved):

salilsub (approved):
Per [this slack thread](https://ngrok.slack.com/archives/C03HT3ELDSQ/p1705955219667889), LGTM

browner12:
Thanks for merging. I can say we've been using this new method in `laravel/settler` and it's working just fine.

PR #575, opened 2024-01-31
Add guide: Ingress to Kubernetes apps deployed on Spectro Cloud Palette

joelhans:

PR #577, opened 2024-02-01
Fix Integrations Hub page for Spectro Cloud

joelhans:
I mis-configured the Integrations Hub frontmatter in #575, which meant the page didn't show associated how-to guide. This PR should fix that.


stmcallister (approved):
That's right! Frontmatter is always picky. Thanks for fixing! 🎉

PR #580, opened 2024-02-05
Adding links to the setup and installation pages in the dashboard

salilsub:
These pages can detect platform and have explicit directions which include the account's authtoken.


russorat (approved):
lgtm

PR #583, opened 2024-02-06
Adding error page for country restrictions

russorat:
the url for this would be https://ngrok.com/docs/errors/country_restriction


TheConcierge (approved):
LGTM. short and to the point. 
i also agree with not including a list here. it'd be too easy for a would-be attacker to look at it and know which countries are "safe".

PR #587, opened 2024-02-07
automated gen that includes policy module api examples

rkolavo:
There was a fix to prevent json in examples to have html escaped characters that also had a side effect of alphabetizing the keys in the json, which seems fine (maybe even good?) to me


wdawson (approved):
Looks 🦵🇮🇹

wdawson:
We will probably want to merge #594 (which will update this PR) before we merge this one.

PR #591, opened 2024-02-09
Add integration how-to guide: Ingress to apps deployed on Azure Kubernetes Service (AKS)

joelhans:
A new how-to guide for AKS and the ngrok Kubernetes Ingress Controller.

I wasn't 100% sure whether to create a new entry in the Integration Hub for AKS, or adopt an existing Azure/Microsoft entry, so I opted for a new one. Happy to migrate this to an existing container if that makes more sense organizationally!


stmcallister (approved):
Looks great! 🌮 🎉 👍

PR #594, opened 2024-02-09
Remove JWT Validation Module entirely because docs pipeliene doesn't.

wdawson:
The pipeline removed examples but all the other stuff is still there :/ 

This PR removes the JWT Module API remnants


rkolavo:
>I'm not sure if this will get clobbered by future pipeline runs if we merge it. 

yea I think it probably would get overridden by new runs of the pipeline if we merged into the dev branch. I think it makes sense to go to main with this

russorat (approved):
lgtm!

PR #598, opened 2024-02-13
custom-response: update configuration

TheConcierge:
Updating the new configuration method for `custom-response`

image

### DIFF
--- a/docs/http/traffic-policy/actions/custom-response.mdx
+++ b/docs/http/traffic-policy/actions/custom-response.mdx
@@ -24,8 +24,9 @@ If this action is executed, no further actions in the traffic policy will be exe
 | ----------------- |
 | `custom-response` |
 
-| Parameter      | &nbsp; | Description                                                                                                        |
-| -------------- | ------ | ------------------------------------------------------------------------------------------------------------------ |
-| `status_code`  | int    | The status code of the response.                                                                                   |
-| `content_type` | string | The content type of the response body, which will be set on the Content-Type header. Required if `content` is set. |
-| `content`      | string | The body of the response.                                                                                          |
+| Parameter      | &nbsp;                    | Description                                                                                                                                                          |
+| -------------- | ------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


ForgoneReality:
Looks like json under the yaml tab?

TheConcierge:
> Looks like json under the yaml tab?

yes! so, in this specific example, we are sending a blob of type 'application/json' back. the content type is a string of json, but overall the format is still (valid) yaml!

ForgoneReality (approved):
in that case, LGTM!

docs/http/traffic-policy/actions/custom-response.mdx
```diff
@@ -24,8 +24,9 @@ If this action is executed, no further actions in the traffic policy will be exe
...
-| `content`      | string | The body of the response.                                                                                          |
+| Parameter      | &nbsp;                    | Description                                                                                                                                                          |
+| -------------- | ------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `status_code`  | int                       | The status code of the response.                                                                                                                   |
+| `content`      | string                    | The body of the response.                                                                                                                          |
+| `headers`      | Map&lt;string, string&gt; | Headers to be added to the response. If `content-type` is not included in `headers`, this action will attempt to infer the correct `content-type`. |
```
rkolavo, line 31: @nijikokun do we want to explicitly call out that this will infer content-type here?

rkolavo (approved):
looks good to me, but would like input from @nijikokun on one thing

docs/http/traffic-policy/actions/custom-response.mdx
TheConcierge: i was thinking it'd be good to document somewhere. maybe this isn't the place. also maybe we don't need to document it at all, maybe its absence is enough to imply to the user that they don't need it unless they really care

tammybailey (approved):

docs/http/traffic-policy/actions/custom-response.mdx
nijikokun: Yes, this is good to know for our end users

nijikokun (approved):

PR #601, opened 2024-02-15
Updating installation instructions with Win 7/8 warning.

salilsub:


docs/getting-started/index.mdx
```diff
@@ -39,7 +39,7 @@ curl -s https://ngrok-agent.s3.amazonaws.com/ngrok.asc | \
 choco install ngrok
 ```
 
-Run this in an Administrator Command Prompt.
+Run this in an Administrator Command Prompt. Version 3.6.0+ require Windows 10 or higher. If you are using Windows 7 or 8, 3.5.0 is the last supported version of the ngrok agent.
```
russorat, line 42: this doesn't feel like it needs to be in the getting started guide. can we add it to the v2 - v3 upgrade guide instead?

salilsub: I don't know if v2-v3 upgrade guide makes sense. There are multiple versions of v3 which work (3.2.x-3.5.0). This is the page where the [install](https://ngrok.com/docs/agent/#install) link points to (other than the [downloads](https://ngrok.com/download) page)

salilsub: @russorat any thoughts on the above?

russorat: yeah, i think https://ngrok.com/docs/agent/#install makes more sense.

PR #604, opened 2024-02-20
fix: added dark mode to whats new

russorat:


nicolaids (approved):
Dark mode and timeframes!

PR #605, opened 2024-02-20
Fix policy gallery for GETs only

krwenholz:
This needs quoting, otherwise you get the following error:

```
ERROR:  failed to start tunnel: Invalid policy expression.
ERROR:  ERROR: <input>:1:2: undeclared reference to 'getDate' (in container '')
ERROR:   | (getDate != 1)
ERROR:   | .^.
```

Unless I did something wrong!


russorat (approved):
lgtm

PR #607, opened 2024-02-21
better macro examples that do not use size

benjaminchan:
update the examples since the previously chosen examples with size compile but do not evaluate as expected


russorat (approved):
lgtm

russorat:
sorry ben, i think i might have messed this up by force pushing. maybe this change was incorporated into something else?

inconshreveable:
no longer clear if this is relevant after the force push, closing. @benjaminchan if we still have `size()` in some of our traffic policy examples, please open a new PR to update the relevant ones.

PR #609, opened 2024-02-22
Add Basic Auth Traffic Policy Example

nijikokun:
Adds Basic Auth enforcement example to the Rule Gallery. This rule mimics the Basic Auth module.

```yaml
# Block all requests without Authorization Header
# Block all requests without valid username:password base64 encodings
inbound:
  - expressions:
      - |-
        !('Authorization' in req.Headers) ||
        !(
          'Basic dXNlcm5hbWU6cGFzc3dvcmQ=' in req.Headers['Authorization'] || 
          'Basic dXNlcm5hbWU6aHVudGVyMg==' in req.Headers['Authorization'] || 
          'Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==' in req.Headers['Authorization']
        )
    actions:
      - type: custom-response
        config:
          status_code: 401
          content: 'Authorization required.'
          headers:
            'WWW-Authenticate': 'Basic realm="MySite", charset="UTF-8"'
```


docs/http/traffic-policy/gallery.mdx
```diff
@@ -72,3 +72,54 @@ This rule applies rate limiting of `30` requests per second to the endpoint `/ap
 		],
 	}}
 />
+
+## Enforce Basic Auth
+
+This rule mimics the behavior of the [Basic Auth module](/http/basic-auth/). Requests that are missing the Authorization header or contain an invalid authorization value will receive an unauthorized response, prompting the requester to authorize themselves.
```
euank, line 78: I think we should also include the caveat that it's less secure.

The basic auth module stores your password hashed in the database, ensuring ngrok doesn't store the plaintext at rest.
This also has the side-effect of making login attempts slower (since we hash the incoming password too for checking), meaning brute-forcing is harder for an attacker.

Using policies necessitates us storing the password plaintext, and as such is suitable for demo and prototyping uses, but not for most prod uses.

It's probably worth having a big warning blurb in this section of "this stores the password in plaintext and should only be done if you understand the security implications of this" so we don't seem like we're irresponsible with respect to security.

euank (commented):
Looks good!

This does make my security sense tingle though! It's clear it's being stored and used unhashed, so if I saw this on a technical company's websites, I'd lost a lot of trust in their security unless it also had a prominent warning of that fact, so as such, I think we should have a prominent security warning until we have a way to input pre-hashed psaswords.

PR #614, opened 2024-02-27
Add a blurb about port allocation

pshafton:
This is a first pass at the port allocation faq entry. Let me know if you want me to add more detail here.

### DIFF
--- a/docs/faq/faq.md
+++ b/docs/faq/faq.md
@@ -79,3 +79,12 @@ _en-grok_
 
 The first prototype for ngrok was committed on [March 20th,
 2013](https://github.com/inconshreveable/ngrok/commit/8f4795ecac7f92c6b5a8c8970c65f26e5315fe4e).
+
+## Why does port scanning my ngrok domain show other open ports {#open-ports-on-domain}
+
+ngrok uses a shared set of servers to service both HTTP(s)/TLS and TCP traffic. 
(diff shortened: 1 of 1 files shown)


shubcodes (approved):
This is the correct response to the customer query about open ports.

docs/faq/faq.md
```diff
@@ -79,3 +79,12 @@ _en-grok_
 
 The first prototype for ngrok was committed on [March 20th,
 2013](https://github.com/inconshreveable/ngrok/commit/8f4795ecac7f92c6b5a8c8970c65f26e5315fe4e).
+
+## Why does port scanning my ngrok domain show other open ports {#open-ports-on-domain}
+
+ngrok uses a shared set of servers to service both HTTP(s)/TLS and TCP traffic. 
+All HTTP(s)/TLS traffic is serviced on ports 80 and 443. For TCP traffic each
```
krwenholz, line 86: Is it worth pointing out that SNI is what gets traffic routed correctly on those ports? If I recall correctly, we had a similar conversation with our pentesters a few years back to get language clarified in our first report.

krwenholz (approved):
Awesome. Good level of detail. Thanks, Peter.

docs/faq/faq.md
pshafton: I added a section about SNI and protecting traffic.

PR #619, opened 2024-02-29
agent 3.7.0 changelog

bobzilladev:


salilsub:
@bobzilladev Can we also add an entry to the version support policy page? Just so they are in sync.

salilsub (approved):
LGTM

PR #632, opened 2024-03-07
Marking region as deprecated in the docs.

salilsub:


salilsub:
@nicolaids I'm marking the region flag as deprecated (has been deprecated since 3.5.0). This seems to be the first thing we are deprecating in the docs, I'd like your take.

nicolaids (approved):
Seems reasonable!

nijikokun (approved):
🚢

PR #634, opened 2024-03-11
Fix and consolidate the intro Policy examples

krwenholz:
Two things were broken in the example:
-  The `content_type` attribute seems to have been removed in favor of `headers` with `content-type`. I think I got that fixed correctly off of everything.
- The `log` example was out of order and broken. It's also not present in all examples, so I just took it out.

I confirmed the change to the agent config. I haven't and probably don't have time to test _all_ of the others.

### DIFF
--- a/examples/agent-cli/http-traffic-policy.mdx
+++ b/examples/agent-cli/http-traffic-policy.mdx
@@ -11,12 +11,9 @@ inbound:
       - type: "custom-response"
         config:
           status_code: 404
-          content_type: text/plain
           content: not found
-      - type: "log"
-        config:
-          metadata:
-            message: "foobar request was made"
+          headers:
+            content-type: text/plain
   - name: "BazCookieForLargeRequests"
     expressions:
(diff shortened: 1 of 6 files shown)


rkolavo (approved):
thanks for catching the headers thing. What do you mean by log examples was "out of order"?

krwenholz:
Two errors in the original example:

```
ERROR:  failed to start tunnel: Failed to parse the configuration for policy action type 'custom-response'.
ERROR:  yaml: unmarshal errors:
ERROR:    line 1: field content_type not found in type action.customResponseConfig
```

then if you fix that you get

```
ERROR:  failed to start tunnel: Policy action type 'log' is unreachable due to a prior terminating action.
```

I'm pretty sure that's because `custom-response` is a terminal action. (According to the docs, anyway!)

rkolavo:
Oh yupppp, no actions will run after terminal, duh. Cool, thanks!

PR #657, opened 2024-03-27
Add integration how-to guide: ngrok Kubernetes Operator + DigitalOcean

joelhans:
This guide uses `doctl` to deploy a new cluster on DOKS with the Operator pre-installed, and `kubectl` to configure it for ingress to an example app.


docs/integrations/digitalocean/k8s.mdx
```diff
@@ -0,0 +1,281 @@
...
+   After a few moments, `doctl` will let you know it's finished deploying the cluster and adding its credentials to your `kubcetl` config file.
+
+   ```bash
+   ...
+   Notice: Successfully kicked off addon job.
+   ID                                      Name        Region    Version        Auto Upgrade    Status     Node Pools
+   d62a17ca-32e0-4d8b-9260-2d0d1c582939    ngrokker    nyc1      1.29.1-do.0    false           running    ngrokker-default-pool
+   ```
+
+1. You can now verify that the cluster is running successfully by looking at the pods in the new `ngrok-ingress-controller` namespace... wait, `CreateContainerConfigError`?
```
stmcallister, line 63: @joelhans Is this a question that should be answered? 😉

docs/integrations/digitalocean/k8s.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+   ```bash
+   kubectl get pods -n ngrok-ingress-controller
+
+   NAME                                                              READY   STATUS                       RESTARTS   AGE
+   ngrok-ingress-controller-kubernetes-ingress-controller-mannvs9b   0/1     CreateContainerConfigError   0          6m30s
+   ```
+
+   That's actually to be expected at this point. If you run `kubectl describe pods -n ngrok-ingress-controller YOUR_NGROK_INGRESS_CONTAINER`, you'd see the following error: `kubelet            Error: secret "ngrok-auth" not found`. Let's remedy that.
```
stmcallister, line 72: Oh! I see. 😆 While I think the question on the previous line is clever, I think for the tone of this doc, I would take it out and start line 72 with something like. "The `CreateContainerConfigError` is actually to be expected...".

docs/integrations/digitalocean/k8s.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+1. Get your the `AUTHTOKEN` and `API_KEY`, which the ngrok Kubernetes Operator will use to authenticate against your existing ngrok account.
```
stmcallister, line 74: "your the" 😉

docs/integrations/digitalocean/k8s.mdx
```diff
@@ -0,0 +1,281 @@
...
+     AUTHTOKEN: "NGROK_AUTHTOKEN"
+   ```
+
+   The ngrok Kubernetes Operator pod should pick up on the new secret and authenticate properly.
+
+   Your new Kubernetes cluster, managed by DigitalOcean, is all set up—and with the ngrok Kubernetes Operator installed with a single option on the CLI.
+
+## **Step 2**: Deploy an example app on your DigitalOcean-managed cluster {#deploy-an-example-app}
+
+Let's explore how you can apps or services to your DigitalOcean-managed Kubernetes cluster. The DigitalOcean has a repository of [sample Kubernetes apps](https://github.com/digitalocean/kubernetes-sample-apps/tree/master/bookinfo-example), including the [Bookinfo](https://github.com/digitalocean/kubernetes-sample-apps/tree/master/bookinfo-example) app, which we'll use.
```
stmcallister, line 103: "Let's explore who you can **add** apps..."

Is it "DigitalOcean" or "The DigitalOcean"? 🤔 I think it's "DigitalOcean has a repository..."

docs/integrations/digitalocean/k8s.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+:::
+
+1. Head over to the [Kubernetes Clusters](https://cloud.digitalocean.com/kubernetes/clusters/) dashboard on the DigitalOcean console and click on the cluster you created previously.
+
+1. Click the **Marketplace** tab, find or search for **Kubernetes Monitoring Stack**, and click **Install**. Confirm your choice.
+
+   In the background, DigitalOcean uses Helm to create a new integrated deployment for Prometheus, Grafana, and Alertmanager.
+
+1. Create a new domain in the [ngrok dashboard](https://dashboard.ngrok.com/cloud-edge/domains) at `monitoring.NGROK_DOMAIN`.
```
stmcallister, line 190: Won't they also need to create the subdomain with their domain registrar? 🤔

stmcallister (changes requested):
Looks great! Just a few comments and questions before publishing.

docs/integrations/digitalocean/k8s.mdx
joelhans: I got a little too cheeky! Made the change as suggested.

docs/integrations/digitalocean/k8s.mdx
joelhans: They wouldn't have to! I actually tested this and it worked fine. I first created an ngrok domain like `digitalocean.ngrok.app` for the initial deployment, then created another one at `monitoring.digitalocean.ngrok.app` for this observability deployment.

Is that clear the way I've described it? Maybe I could be clearer about what I did, or suggested creating a domain like `monitoring-digitalocean.ngrok.app` instead.

stmcallister: 🤔 Got it. With that context, that makes sense. I was looking at it through the "custom domain" lens. But, we never mention those in the scope of this article. So, I think we're good. Thanks for the explanation!

stmcallister (approved):
Looks good! 🌮 👍 🎉

PR #674, opened 2024-04-11
Fix redirect configuration

earl-ngrok:
change type to redirect instead of custom-response.


nijikokun (approved):

nijikokun:
@earl-ngrok just need to run `pnpm run fmt`

PR #680, opened 2024-04-19
Mandy/site to site apis

DevMandy:
Added a new guide for site-to-site connectivity to APIs


shubcodes (approved):
This was great -- would like to see an additional copy of this exact guide that just does it with HTTP and has examples for IP restrictions etc.

PR #682, opened 2024-04-19
Add `edit` links to error pages

krwenholz:
While most of the page is generated by internal tooling, this allows our community to suggest changes!

This does include a number of API example updates and changes, but I don't think we can exclude those from the automation right now. Let me know, @russorat !


russorat (approved):
lgtm

PR #683, opened 2024-04-20
Fix the links in flask.md

JxxIT:
This directly links to the configuration variables in the docs. The link to the stackoverflow post was changed, so I added it directly.


salilsub (approved):
LGTM

PR #686, opened 2024-04-24
agent changelog for 3.9.0

bobzilladev:
Agent changelog for `3.9.0`

### DIFF
--- a/docs/agent/api.md
+++ b/docs/agent/api.md
@@ -411,3 +411,39 @@ GET/api/requests/http/:request_id
     }
 }
 ```
+
+## Agent Status
+
+Get status and metrics about the running agent
+
+### Request
+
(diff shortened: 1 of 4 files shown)


salilsub (approved):
We should also work on docs for the new endpoint, although that doesn't need to block the changelog.

bobzilladev:
> We should also work on docs for the new endpoint

Good point, I took a stab at it, wdyt? https://github.com/ngrok/ngrok-docs/blob/fef90737f29442f9679be6fe4ce2eeddb1a2985e/docs/agent/api.md#agent-status

salilsub:
I think we might need some additional detail in the docs page, but we can also deal with that later. I don't think that adding those agent api docs should block the changelog or the release.

PR #696, opened 2024-05-02
Fix the wording on Add Headers

pshafton:
Update to indicate we are adding headers `to` the request or response rather than `from` the request or response


russorat (approved):
lgtm

Issue #734, opened 2024-05-15
Using ngrok with Docker compose -- clarify need for host.docker.internal in ngrok.yml on Mac


justngrok:
Need to update "Using ngrok with Docker Compose" to clarify that the defined tunnels in agent config needs to also use the host.docker.internal hostname for upstream.

https://ngrok.com/docs/using-ngrok-with/docker/

In the "Using ngrok with Docker", we do call out you have to use host.docker.internal for MacOS

This also applies to the upstream addr: for the ngrok.yml when following the Docker Compose instructions on mac.

> Note: for MacOS or Windows users, the --net=host option will not work. You will need to use the special url host.docker.internal as described in the [Docker networking documentation](https://docs.docker.com/desktop/mac/networking/#use-cases-and-workarounds).

Example:
```
tunnels:
  test:
    proto: http
    addr: http://host.docker.internal:80
```

This will allow docker to start, but returns ERR_NGROK_8012 (upstream error) when accessing the endpoint.
```
tunnels:
  test:
    proto: http
    addr: 80
```

justngrok:
Title says Mac, but this applies to windows too.

PR #739, opened 2024-05-17
fix: adding note about using internal addresses in ngrok config files

russorat:
closes https://github.com/ngrok/ngrok-docs/issues/734


justngrok (approved):
Looks good.

PR #749, opened 2024-05-23
update docs with new variables

benjaminchan:


rkolavo (approved):
Ben, this is really great — thanks for the thorough effort that this was

PR #770, opened 2024-06-04
add new doc changes for url-rewrite

KristopherPaulsen:
image

Bikeshedding and wordsmithing welcome.


rkolavo:
Hmmm I think we need to indicate to users more clearly that the Named Variables shown are for nginx format and use the ngrok variables/cel expressions for ngrok format

rkolavo (approved):

PR #776, opened 2024-06-06
agent release 3.10.1

bobzilladev:


docs/agent/version-support-policy.mdx
```diff
@@ -44,7 +44,7 @@ Extended support beyond LTS is available to customers on a commercial basis. It
 
 | Support Status | Release | Release Date | Latest | Active Support | End of Life |
 | -------------- | ------- | ------------ | ------ | -------------- | ----------- |
-| Active         | 3.10x   | 2024-05-23   | 3.10.0 | 2025-05-23     | 2025-07-23  |
+| Active         | 3.10x   | 2024-06-06   | 3.10.1 | 2025-06-06     | 2025-08-06  |
```
salilsub, line 47: I don't think we would be changing the release, active support or EOL date. Those will be based on the date the minor version was originally released.

```suggestion
| Active         | 3.10x   | 2024-05-23   | 3.10.1 | 2025-05-23     | 2025-07-23  |
```

salilsub (changes requested):
We should leave those dates as is since we won't delay deprecation every time we release a new patch.

Abdiramen (approved):
:shipit:

salilsub (approved):
LGTM

PR #794, opened 2024-06-14
Audition alternative body typefaces

aaronshekey:
We've had feedback over the years that the legibility of our docs could be improved. Let's consider switching to a different typeface for better readability.

This PR simply hotlinks to Google Fonts as an import, and provides a few options. You can switch to various typefaces by changing the font in the topbar.


aaronshekey:
Closing since we've made the decision to set our body type in Nunito Sans #851 🎉

PR #798, opened 2024-06-18
Update changelog.md to include port 80 for CRL address change

justngrok:
and the port 80 as multiple customers confused and are setting domain with 443.


ctindel (approved):
LGTM

PR #800, opened 2024-06-20
Update error 108

kaganjd:
Add `ngrok start --all` to error message, remove dashboard link


russorat:
these error messages are actually auto generated from our platform code. i can make this change there and then it will propagate over

kaganjd:
ah, gotcha- thanks! ok to make edits in `docs/errors/details/` though?

PR #805, opened 2024-06-24
updates for variables

benjaminchan:
- update timestamp variables for conn
  - update examples using old conn timestamps
- add timestamp variables for req/res
- add endpoint variables 
- update req.url.host
- add req.url.port, req.url.raw
- add conn.client_port in tcp/tls
- update http request header templates to include these
- update header macro descriptions to no longer require canonical format


docs/http/traffic-policy/expressions/macros.mdx
```diff
@@ -7,14 +7,14 @@ also be used in policy expressions. For convenience, the following custom macros
 
 | Name                                                                       | Return Type | Description                                                                                                                                       |
 | -------------------------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
-| [`hasReqHeader(string)`](#hasreqheaderstring)                              | bool        | Returns true or false if the provided header key is present on the request. Header keys must be written in canonical format.                      |
```
benjaminchan, line 10: should we keep these?  i removed as not necessary anymore given we preprocess the keys to lower case, but maybe it's better user experience to provide narrower instructions

rkolavo: yeah I say remove, type as crazy as you want, dear users

docs/http/traffic-policy/expressions/variables.mdx
```diff
@@ -891,6 +891,78 @@ Start date and time when the certificate becomes valid.
...
+	}}
+/>
+
+### `endpoint.port`
+
+The port for this endpoint.
+
+<ConfigExample
+	config={{
+		expressions: ["endpoint.host == 443"],
```
rkolavo, line 942:
```suggestion
		expressions: ["endpoint.port == 443"],
```

rkolavo (approved):

PR #810, opened 2024-06-26
add action results variables and update other variables

benjaminchan:
- action results
- add templates for vars
- removed byte count, timestamp vars that don't apply to various phases
- add endpoint vars to tcp/tls
- add aliased variables for cert extension and cookies
- remove outbound log action for tcp/tls


docs/traffic-policy/variables/endpoint.mdx
```diff
@@ -0,0 +1,73 @@
+import ConfigExample from "../../../src/components/ConfigExample.tsx";
```
rkolavo, line 1: Instead of breaking these up by namespace I was thinking more by "phase" (as in on_tcp_connect, on_http_request, on_http_response). What do you think?

benjaminchan: I was thinking namespace to be able to have different reference tables for tcp vs tls endpoints (since conn.tls vars isn't useful for tcp). Open to building up by phase too, might be missing how you'd want to use them

rkolavo: nah you're right.... tls has different vars than tcp

rkolavo (approved):

PR #813, opened 2024-06-26
fix copy-text and make docs ngrok rock!

KristopherPaulsen:
Slight fixes to copy-text


rkolavo (approved):
every time I read these examples I get rockapella stuck in my dang noggin

PR #830, opened 2024-07-10
fix: adding info about our agent urls and ips

### LINKED ISSUE #753: Add information on addresses used by ngrok Agent to connect
Start a discussion if we should list the multiple addresses the Agent will use to connect to ngrok.

We assume that 443 and 80 outbound are open, but this isn't the case in larger organizations. They also have to whitelist outbound domains, etc.

Maybe add something here?
https://ngrok.com/docs/guides/running-behind-firewalls/

Agent Connection addresses
CRL addreses
Update (equinox) address.

russorat:
closes #753


justngrok (approved):
lgtm

PR #844, opened 2024-07-15
switch to pnpm v9

cody-dot-js:
### Why

Let's upgrade to the latest version of pnpm for performance, stability, and feature enhancements.

### How

Leverage `direnv`, `nvm`, and `corepack` to install the required tooling. Update the readme for users.

Note: we shouldn't merge this until the monorepo has switched to pnpm v9 as well (for the docs pipeline to work properly)


.envrc
```diff
@@ -1 +1,26 @@
-use flake
+# Check if nvm is installed, install if not
```
jtszalay, line 1: fwiw. This will make anyone using nixos unable to use the envrc (unless they preinstall nvm it elsewhere). It can't just curl and install (easily). So that'll impact @nikolay-ngrok @euank @jrobsonchase @TheConcierge if that matters.

cody-dot-js: how often do they work w/ the docs?
is there a way to have multiple `.envrc`? something like

```
direnv allow .envrc.nix
```

jtszalay (approved):

PR #850, opened 2024-07-17
cookie and port variables fixes

benjaminchan:
- fix cookie macros types and update description (to account for multiple cookies of same name) and example
- fix cookie variables types and update description (to account for multiple cookies of same name)
- add res.cookies[k][i].unparsed_attribute variable
- update req.url.port description (to call out what happens when port is implied)
- update req/res.cookie, req.url.host descriptions for http header templates variables


docs/http/_http_header_templates.mdx
```diff
@@ -85,7 +85,7 @@ The source of this data is subject to change. It is currently provided by [MaxMi
 | `${req.content_type}`            | The media type set in the Content-Type header for this request as a string.                                                                                                                     |
 | `${req.content_type.parameters}` | The parameters set in the Content-Type header as a key value map.                                                                                                                               |
 | `${req.content_type.raw}`        | The Content-Type header for this request as a string.                                                                                                                                           |
-| `${req.cookies}`                 | The key value map of HTTP cookie objects provided in the request.                                                                                                                               |
+| `${req.cookies}`                 | The key value map of HTTP cookie objects included in this request. If there are multiple cookies with the same name, they will be ordered as specified in the Cookie header.                    |
```
rkolavo, line 88: Is it the `Set-Cookie` header?

benjaminchan: Should be `Cookie` since that's the one used by clients in requests, whereas `Set-Cookie` is used by servers in responses

rkolavo: yup I'm losin it

rkolavo (approved):

PR #862, opened 2024-07-26
3.14 release: add support for `traffic_policy`

TheConcierge:
The agent has been updated to support sending policy as a string to the backend. This was added for both ephemeral tunnels and edge configurations with the API.

These changes are part of our ongoing effort to rename policy phases.


TheConcierge:
These are proposed docs changes for the agent release going out on Thursday

docs/agent/changelog.md
```diff
@@ -6,6 +6,12 @@ title: Changelog
 
 ## v3
 
+### ngrok Agent 3.14.0 - \[2024-08-01\]
+
+- Added `--traffic-policy-file` flag that accepts Traffic Policy configuration for HTTP, TCP, or TLS traffic.
```
russorat, line 11: are these also updated in https://ngrok.com/docs/agent/cli/ and https://ngrok.com/docs/agent/config/ ?

russorat (approved):
LGTM. might need to also update the docs for the agent cli and config.

docs/agent/changelog.md
TheConcierge: no but they will be. since the old fields still work for the time being, there is no absolute hurry. there is possibly an incentive to hold off, as these new names aren't present for old agent version. writing these doc changes was my next goal. However, if people think we should just do it as part of this PR, I can add it in.

TheConcierge: resolving because these doc changes either already existed or were added to the PR

nijikokun (approved):

PR #866, opened 2024-07-30
Update _err_ngrok_3200.md

justngrok:
Added information for end users following links, that don't have ngrok accounts.

Would like to consider temporarily adding information for Aqua Protocol tickets, would like to discuss this before PR is approved.


russorat (approved):
lgtm

PR #871, opened 2024-07-30
Update limits.md

justngrok:
Added Interstital bypass examples for common libraries. Can these be displayed a better way?

### DIFF
--- a/docs/guides/limits.md
+++ b/docs/guides/limits.md
@@ -77,4 +77,42 @@ If the interstitial interferes with your project, you can bypass it in a few way
 - Change your user agent by setting the `User-Agent` header to something non-standard, such as `MyApp/0.0.1`, to bypass the warning.
 - Upgrading to any of our paid plans will also bypass the warning.
 
+### Code Examples
+Axios
+```
+axios.get(url, { 'headers': { 'ngrok-skip-browser-warning': '1' } })
+  .then((response => {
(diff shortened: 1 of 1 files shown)


docs/guides/limits.md
```diff
@@ -77,4 +77,47 @@ If the interstitial interferes with your project, you can bypass it in a few way
 - Change your user agent by setting the `User-Agent` header to something non-standard, such as `MyApp/0.0.1`, to bypass the warning.
 - Upgrading to any of our paid plans will also bypass the warning.
 
+### Code Examples
+
+Axios
+
+```
+axios.get(url, { 'headers': { 'ngrok-skip-browser-warning': '1' } })
```
kaganjd, line 85: fwiw, i tried this method by adding a js script to an index.html file and was not successful. could/should we include a full working index.html file here?

kaganjd: could also be nice to add or build a basic chrome extension since i have had customers say this is the only way that will work for them (app is server only)

i also was not able to do this with nextjs because of server-side rendering or something- should we call out this limitation somewhere?

russorat: this post has the chrome extension https://stackoverflow.com/a/78128230/17282727

russorat (approved):
lgtm. we can add more examples and the browser extension later.

PR #873, opened 2024-07-30
feat: add headings to common 3200 errors

nijikokun:
Make it easier to scan the 3200 common errors by adding some bold headings.


russorat (approved):
lgtm

PR #890, opened 2024-08-09
Update policy to traffic_policy on HTTP/TLS/TCP 

joelhans:
I noticed that after #862, some instances of example config files were changed from `policy` to `traffic_policy`, but that hadn't reached the Traffic Policy-specific doc ([example for HTTP](https://ngrok.com/docs/http/traffic-policy/?cty=agent-config)) for each type of endpoint.

This PR updates those three examples to update the naming. And fixes what I believe to be malformed indentation in the TCP example.

### DIFF
--- a/examples/agent-config/http-traffic-policy.mdx
+++ b/examples/agent-config/http-traffic-policy.mdx
@@ -3,7 +3,7 @@ tunnels:
   example:
     proto: http
     addr: 80
-    policy:
+    traffic_policy:
       inbound:
         - name: FooBarParamNotFound
           expressions:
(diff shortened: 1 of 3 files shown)


joelhans:
Hi @TheConcierge and @rkolavo! I'm not sure how review requests do/should work here, or if either of you two are the right person to ping here, so just let me know—thanks!

rkolavo (approved):

TheConcierge:
That's my bad. Good catch, Joel!

TheConcierge (approved):

joelhans:
@TheConcierge Happy to help! Just noticed it as I was investigating the changelog ahead of working on the next newsletter.

PR #891, opened 2024-08-09
Add 2nd docker-compose example

stmcallister:
As per feedback in the Product interlock meeting, this PR adds a second example in the Using ngrok with Docker documentation for running the `ngrok/ngrok` container inside a `docker-compose.yml` file without using an `ngrok.yml` file to define a tunnel. This example shows how to use the `host.docker.internal` url in an `ngrok http` command.


joelhans (approved):
LGTM!

PR #893, opened 2024-08-12
Update ERR_NGROK_8012 documentation and add details file

earl-ngrok:


russorat (approved):
the main error file will get overridden the next time we generate the docs, but the _ file looks good.

PR #895, opened 2024-08-13
Update ERR_NGROK_3208 documentation and add details file

earl-ngrok:


russorat:
What is the reasoning for removing the section from the error message? if we want to do that, we will need to update the autogenerated error message.

earl-ngrok:
I moved the content to the "For Website Visitors" section thinking it fit better there, but I see the value in keeping it in the error message too. We can include it in both places and leave the original error message untouched if this works for you.

russorat (approved):
LGTM

PR #896, opened 2024-08-13
fix: Update docs for kubernetes CRDs

jonstacks:
Replace the embedded struct ngrokApiCommon with the actual properties it provides on the resources.


nijikokun (approved):
🙌

PR #899, opened 2024-08-21
Turn Full Example Config into a Fully Documented Default Configuration File

nijikokun:
Many users copy and paste the full example blindly into their configuration file which leads to issues. This change converts the full example into a fully documented configuration file similar to other tools which should lead to forcing users to read each section before enabling functionality. In addition to this, there are numerous improvements which include linking to configuration fields for tunnels, including links to dashboard for obtaining tokens and keys, etc.


russorat (approved):
reminds me of the apache config file :) LGTM

PR #904, opened 2024-08-23
Update agent config documentation

KristopherPaulsen:


KristopherPaulsen:
discovered more items that need love and attention

docs/agent/config/v2.mdx
```diff
@@ -1,43 +1,11 @@
 ---
-title: Configuration File
+title: "Version 2"
```
KristopherPaulsen, line 2: This is "almost" a 1-1 copy of the previous configuration file, but now instead under v2

PR #906, opened 2024-08-23
docs: scheme -> url

warrn:


docs/agent/config.mdx
```diff
@@ -493,7 +493,7 @@ tunnels:
 ngrok start httpbin
 ```
 
-Each tunnel you define is a map of configuration option names to values. The name of a configuration option is usually the same as its corresponding command line switch with hyphens (`--host-header` becomes `host_header:` in the configuration file and `--scheme` becomes `schemes`). Tunnels can define a specific `proto` or use labels to dynamically connect to one or more matching ngrok Edges. All tunnels must define a specific `addr` that tells the agent where to send the traffic. Other properties are available and many are protocol-specific.
+Each tunnel you define is a map of configuration option names to values. The name of a configuration option is usually the same as its corresponding command line switch with hyphens (`--host-header` becomes `host_header:` in the configuration file and `--url` becomes `url`). Tunnels can define a specific `proto` or use labels to dynamically connect to one or more matching ngrok Edges. All tunnels must define a specific `addr` that tells the agent where to send the traffic. Other properties are available and many are protocol-specific.
```
KristopherPaulsen, line 496: I think this chunk will conflict with my changes in https://github.com/ngrok/ngrok-docs/pull/909, so you can probably safely remove it 🖖

KristopherPaulsen (approved):
Looks good, one small request to avoid conflicts with my incoming branches 🖖

PR #907, opened 2024-08-23
Fix command to start tunnel with mTLS and add formatting

DevMandy:
Changed `client-crt.pem` to `server-crt.pem` in the three new guides that use mTLS. Also formatted a `201` that was missing backticks. Simple PR.


joelhans (approved):
Lovely—that should address the concern we chatted about before. Nice catch on those closing brackets, too. LGTM!

PR #915, opened 2024-08-29
fix: remove server addr from v3 docs

russorat:


russorat:
closed as out of date

PR #927, opened 2024-09-06
Add tip+link to Office Hours from quickstart

joelhans:
Given that our ideal audience for Office Hours is folks who have recently started with ngrok, this doc feels like the perfect place to plant the seed of another place where they can get some help.


russorat (approved):
lgtm

PR #930, opened 2024-09-10
feat: upgrade docusaurus

russorat:


joelhans (approved):
Good call on updating those links to agent configs.

The deploy preview looks solid. Nothing out of the ordinary happening. I peeked at the [Docusaurus changelog](https://docusaurus.io/changelog) and it doesn't seem like there are any breaking changes or things we should worry about updating alongside pinning the new version.

PR #938, opened 2024-09-13
fix: /docs/network-edge/pops redirect loop

euank:
Before this, going to https://ngrok.com/docs/network-edge/pops/ would redirect loop.

The reason for this is twofold:

1. The match should have been exact, since the redirect was a prefix of itself
2. The redirect check at the bottom checked href ('https://ngrok.com/docs/....') against a path ('/docs/') to see if they were the same, and so they never were.

This updates that check as well.


russorat (approved):
lgtm! thanks!

PR #944, opened 2024-09-18
Update how-to-set-up-a-custom-domain.md

pkinchla:
starting a tunnel with a custom domain should use the `domain` flag rather than the `url` flag


russorat:
@pkinchla thanks for the issue. in the latest agent v3.16.0, the flag has been changed to `--url`. the old style still works though.

pkinchla:
@russorat Thanks for the info. Much appreciated.

PR #977, opened 2024-10-10
Document how to run the agent as a job in k8s to diagnose the network

alex-bezek:
The ngrok diagnose command is helpful for people first trying out ngrok on corporate networks to see if we are blocked by their network/security teams. 

Within a kubernetes cluster, we can have the same issue. Often times we'll give people dig, nslookup, ping, or curl requests to test their network. This documents how to run the ngrok agent in a docker container in the cluster and have it report the diagnose command. 

Longer term, this will likely be incorporated into the ngrok-operator helm chart, but it might be useful to have here as well for other situations.


kaganjd (approved):
this is awesome- thank you!

PR #980, opened 2024-10-14
add-server-addr is deprecated, so we need to update the documentation…

KristopherPaulsen:
Okay, technically it's "aliased", but we're going to deprecate it. I've updated to use `add-connect-url`.


cody-dot-js (approved):
🦐

andrewryno (approved):

PR #985, opened 2024-10-16
Turn off bracket spacing rule for "variables" in docs/guides

joelhans:
In code or configuration blocks in our guides, @DevMandy and I often use "variables" that users must replace with a real value, like replacing `{YOUR_NGROK_DOMAIN}` with `example.com`.

We recently noticed the Prettier CI job adds spaces between the bracket and the variable, like `{ YOUR_NGROK_DOMAIN }`. For the sake of consistency, we'd like to not have these spaces automatically injected into our code/config samples, as that messes up how we refer to them throughout our guides. 

This setting disables the bracket spacing, which gets me and @DevMandy what we want... are there any objections to disabling this fix? Places where disabling it might cause issues?


DevMandy (approved):
I approve but we should get more buy-in before pushing. :)

russorat (approved):
looks good!

joelhans:
FYI for anyone checking this PR out after merge: I revised my edit to the Prettier config so that it only affects `.md/.mdx` files in the `docs/guides/` directory so I'm not messing with any files outside of the immediate concern from me and @DevMandy. We can always reassess later to change this on *all* docs.

PR #986, opened 2024-10-16
Switch --url back to --url when the argument is, in fact, a domain...

KristopherPaulsen:
I got a little overly aggressive with my original regex replace for `--domain` to `--url`. I replaced some items where it _should_ be `domain` (`--url` doesn't make sense for creating a _domain_).


nijikokun (approved):
looks like this oopsie is about to be fixed, lgtm

PR #996, opened 2024-10-22
fix typo in traffic policy docs

jmhertlein:


jmhertlein:
(to reviewers: it suggested you two, so i clicked ya both)

TheConcierge (approved):
oopsie

PR #1000, opened 2024-10-22
Added redirects.

blyonng:
Cleaning up some SEO 404 broken links.


DevMandy (changes requested):
Remove the trailing `/` after each URL or it will only work if they include the `/`.

blyonng:
> Remove the trailing `/` after each URL or it will only work if they include the `/`.

Updated

russorat (approved):
LGTM

PR #1012, opened 2024-10-29
fix: update wordpress docs

russorat:
latest wordpress release doesn't require any of the custom editing to work with ngrok.


kaganjd (approved):
neat! small q: is it worth removing the tip at the top about moderate to advanced wordpress knowledge?

PR #1014, opened 2024-10-30
Mandy/api gw getting started

DevMandy:


docs/guides/api-gateway/get-started.mdx
```diff
@@ -0,0 +1,644 @@
...
+step-by-step instructions for integrating with Auth0, but you can configure JWT validation for any OAuth provider. Check
+out our [integrations guides](/docs/integrations/) for specific examples.
+
+#### Define your API in Auth0
+
+1. Login to your [Auth0 Tenant Dashboard](https://manage.auth0.com/dashboard):
+1. Select **Applications > APIs**
+1. Click **+ Create API**
+1. Name your API whatever you'd like
+1. Add your ngrok domain as the identifier
```
joelhans, line 555: Might be helpful to call this out specifically with a variable.

joelhans (changes requested):
Hey @DevMandy! Amazing work here! I've tested all your code/rules, and everything works as expected, which is **awesome**. Great demos.

Aside from these, just a few higher-order things to think about. These could definitely be a follow-up improvement and not a blocker for you!

1. I think we need some type of conclusion here. The way it ends is a bit... arresting? Some links to other documentation, or even the API-G+K8s guide, would be great.
2. You use a few different variables and formats to discuss the user's ngrok domain. I'd love if you just stuck with `{YOUR_NGROK_DOMAIN}` throughout, without the `ngrok.app` part, so that folks can choose which path works best for them and your variable usage remains consistent.
3. I'd love if our guides were consistent with terminology. From my understanding of the [spec](https://www.notion.so/ngrok/Traffic-Policy-Module-742cb71e635f481b8f7179507aa34fca#78b7b77492aa4ba9a2325f5026007d52), the correct term for the configuration in the `policies.yaml` file is **rules**.

docs/guides/api-gateway/get-started.mdx
```diff
@@ -0,0 +1,644 @@
...
+                              0       1       0.00    0.00    0.00    0.00
+
+HTTP Requests
+-------------
+
+15:00:14.035 CDT GET /cards                     200 OK
+
+```
+
+Change the header value to `2` and run the command again. You should see this request hit the `https://v2.api.internal`
```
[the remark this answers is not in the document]
DevMandy: 😄

docs/guides/api-gateway/get-started.mdx
```diff
@@ -0,0 +1,644 @@
...
+---
+on_http_request:
+  - expressions: []
+    name: Add rate limiting
+    actions:
+      - type: rate-limit
+        config:
+          name: Only allow 30 requests per minute
+          algorithm: sliding_window
+          capacity: 30
```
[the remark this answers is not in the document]
DevMandy: Sure thing!

docs/guides/api-gateway/get-started.mdx
```diff
@@ -0,0 +1,641 @@
...
+#### Test traffic policies
+
+Run the following command to hit the `/cards` route:
+
+```bash
+curl -X GET https://{YOUR_NGROK_DOMAIN}/cards
+```
+
+Examine the terminal where you started the `https://hostA.api.internal` endpoint on port `8001`, and you can
+see that the request hit the correct endpoint. Repeat the steps for the `/payments` endpoint and confirm the request
```
joelhans, line 407: Saw this when helping with this rule earlier. Let's make doubly sure we're only using **endpoint** alongside cloud/internal to avoid any possible confusion.

```suggestion
see that the request hit the correct endpoint. Repeat the steps for the `/payments` route and confirm the request
```

joelhans (changes requested):
One more thing I missed before!

docs/guides/api-gateway/get-started.mdx
DevMandy: I saw your one comment about that and then updated `endpoints` to `routes` everywhere in the guide. Thanks for flagging that.

joelhans (approved):

docs/guides/api-gateway/get-started.mdx
```diff
@@ -0,0 +1,641 @@
...
+          url: https://v2.api.internal
+          binding: internal
+```
+
+:::note
+You must always include the `forward-internal` action as the final step in your policy file since once it executes,
+no subsequent policy rules will be executed. Additionally, you always need a terminating policy rule.
+In this case, the traffic forwards to `https://v2.api.internal` if the header is set to a value other than `1` or `2` or
+if the header isn't present in the request.
+:::
```
rkolavo, line 262: Here and in the other notes about having to end with a `forward-internal` action is not *quite* accurate. 

The traffic policy validation (for cloud endpoints) forces you to have some action implemented that is guaranteed to do _something_ with traffic on your endpoints. This could be `forward-internal` but this could also be `deny` or `custom-response`.

DevMandy: That's totally true! I'll update these warnings to be more technically accurate. Thanks for the feedback, Rachel.

PR #1015, opened 2024-10-30
Remove beta billing from internal endpoints

joelhans:
Was asked by @samcrichard to remove this, as I grabbed this info when launching the internal endpoints.

This removes the beta verbiage and replaces with some copy from Sam.


russorat:
i'm pretty sure all the endpoints and policy stuff is still under dev preview.

inconshreveable (approved):

PR #1021, opened 2024-11-04
Docs: change the download URL via find-and-replace

blyonng:
FROM: `https://ngrok.com/download`
TO: `https://download.ngrok.com`


cody-dot-js (approved):
🦐

PR #1022, opened 2024-11-07
Fix the oauth.scopes, which is incorrectly named

alexdlaird:
Fix the oauth.scopes, which is incorrectly named.


russorat (approved):
lgtm

PR #1034, opened 2024-11-14
Update K8s API gateway guide for new ngrok-operator

joelhans:
Making a few QOL improvements and aligning with new product updates in ngrok/ngrok-operator:

1. Made the example API create its own namespace, removing the `kubectl create namespace prod` step
2. Updated the `helm` command to use the new `ngrok/ngrok-operator` chart! 🥳 
3. Moved the operator to the `ngrok-operator` namespace now that they can work in _different_ namespaces.
4. Moved to `on_http_request` instead of `inbound` now that phase-based naming works in the operator.
5. Fixed the `curl` loop example to use `POST`, which will better show the rate limiting in action.


DevMandy (approved):
I retested the whole thing and it's golden!

PR #1035, opened 2024-11-15
Adding full forwardInternal example

justngrok:


justngrok:
typescript isn't my strong suite, so the example may need help.

docs/http/traffic-policy/gallery.mdx
```diff
@@ -156,3 +157,18 @@ Prevent obsolete and potentially vulnerable browsers, SDKs, or CLI tools like `c
...
+[URL Rewrite](/docs/http/traffic-policy/actions/url-rewrite/) Traffic Policy Actions to route requests to differnt upstream ports based on the URL path of the request.
+This allows us to use one domain to route to multiple services (ports) based on the request path sent in each request.
+
+https://mypublic.ngrok.app/... --> http://localhost:80/...
+https://mypublic.ngrok.app/service1/... --> http://localhost:81/...
+
+In this example, we are creating a "public" endpoint using an ngrok reserved domain "mypublic.ngrok.app", an "internal" endpoint (only accessible inside your acccount) "service1.internal",
+then using the forwardInternal action to forward requests that have "/service1" in the path to a different port.
+Finally, we remove the added path from the request before sending it to the upstream service using the URL Rewrite action. This path then does not have to exist on the upstream service, creating a "virtual" "service1" path on the public ngrok endpoint.
```
S3Prototype, line 172:
```suggestion
### Routing requests based on request path

You can route traffic from one domain to multiple services (ports) based on the path sent in each request.

In your Agent [configuration file](/docs/agent/config/), you can modify both the [forwardInternal](/docs/http/traffic-policy/actions/forward-internal/) and
[URL Rewrite](/docs/http/traffic-policy/actions/url-rewrite/) Traffic Policy Actions to route requests to different upstream ports based on the URL path. For example you may want to route public traffic to a URL such as `https://mypublic.ngrok.app/...` to `http://localhost:80/...`.

The following example:
- Creates a public endpoint using an ngrok reserved domain – "mypublic.ngrok.app"
- Create an internal endpoint – "service1.internal"
- Uses the `forwardInternal` action to forward requests that have `/service1` in the path to a different port.
- Removes the added path from the request before sending it to the upstream service using the URL Rewrite action.
  - This path does not have to exist on the upstream service,  as a virtual `service1` path will exist on the public ngrok endpoint.
```

S3Prototype (approved):
After some fixes, the PR looks good!

PR #1036, opened 2024-11-15

Alice-Lilith:


blyonng:
@Alice-Lilith is there any possibility that auto-generating these will lead to deleted or moved pages? If so, in a followup ticket, is there a way we can flag any auto-generated moved or deleted URLs, so that we can add redirects?

russorat:
closing this in favor of #1058

PR #1056, opened 2024-12-03
Update index.mdx

ZaneBartlett1:
Updated documentation so that setup for Macs work

Notes: 
* I had ngrok installed, ran the command, got the message "Press ctrl+U to update", updated, ran the command again, then got errors like "zsh: command ngrok not found" until I uninstalled. After looking into it, and trying to myself, using `brew install ngrok/ngrok/ngrok` only works if you tap the repo first. This update adds that to the setup. 

* This is updating the documentation, but I'm not sure how your https://download.ngrok.com/mac-os works, and also has this issue

* Lastly, MacOS isn't my daily driver, I hope this fix is correct, but you may want to refer to someone with more experience. All I know is that trying the set up instructions didn't work as written, and this fixed it, and makes sense for why as I understand it

### DIFF
--- a/docs/getting-started/index.mdx
+++ b/docs/getting-started/index.mdx
@@ -17,6 +17,12 @@ own application running locally.
 <Tabs groupId="os" queryString="os">
 <TabItem value="macos" label="Mac OS" default>
 
+Add our tap so Homebrew knows where to find the official ngrok formula. This will be more up to date than what may be in homebrew's default repository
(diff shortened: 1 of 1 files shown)


russorat:
@ZaneBartlett1 thank you for the PR! we're discussing internally if we should continue to use our own tap or use the one in homebrew-core. I'll get back to you soon.

PR #1062, opened 2024-12-05
Make it so the deploy workflow doesn't clobber the rss feed file in t…

alex-bezek:
…he bucket

We found that the rss feed will 404 sometimes depending on how recently a deploy job happened. This makes it so a deploy won't remove the RSS files


nijikokun (approved):
lgtm

deroine (approved):
Thanks for tracking this down, nice fix!

PR #1066, opened 2024-12-09
feat: forward-internal url supports cel interpolation

nijikokun:


rkolavo:
how do we get the little badge thing like other config fields? 
image

russorat (approved):
lgtm

PR #1067, opened 2024-12-09
fix: inbound -> on_http_request for forward-internal example

nijikokun:
we use `on_http_request` now :)


rkolavo:
oh nice catch! I didn't even see this one — but there's one more in the Behavior section (I don't think I can do a suggestion in a separate file 😭 )
image

russorat (approved):
lgtm

PR #1011, opened 2024-10-28
New Traffic Policy Documentation

nijikokun:
This is the PR for the new Traffic Policy documentation. Goals is to move Traffic Policy out of the Endpoints and into it's own dedicate space.

PR #1068, opened 2024-12-09
Add circuit-breaker action to new Traffic Policy docs

joelhans:
I noticed that after #1011, the circuit breaker action docs disappeared, so this is my hopefully-successful attempt to port over the existing file into the new format.


docs/traffic-policy/actions/index.mdx
```diff
@@ -44,6 +44,16 @@ actions:
     categories:
       - security
 
+  - type: circuit-breaker
+    name: Circuit Breaker
+    description: Maintain system reliability by rejecting requests when error rates exceed defined thresholds.
+    phases:
+      - on_http_request
+      - on_http_response
```
rkolavo, line 52: I think just `on_http_request` here

joelhans: Yep you're right—good catch & thank you!

nijikokun (approved):

PR #1069, opened 2024-12-09
v2.0 of customer's customer guide

joelhans:
This PR answers a few questions we didn't get around to with the first round. I also some have additional feedback from @russorat to scope and add to a `v2.5` change in January.

@DevMandy Could you take a look at the changed sections overall like you did with v1?

@justngrok Could you peek at the **Can I set up deep packet inspection or observe what data transmits through ngrok?** to see if I missed anything?


DevMandy (approved):
lgtm

PR #1070, opened 2024-12-10
Avoid running deploys in parallel

euank:
The `aws s3 sync --delete` we're using can conflict with itself if run in parallel.

Specifically, if one action is trying to deploy a new version of docs, and at the same time another action is trying to deploy a different, also new version of docs, they can race such that i.e. we get "index.html" from one, but "main.<hash>.js" from the other, and for `main.<hash>.js` from one to be deleted by the other run

In order to avoid that happening, add github's concurrency control to stop any previous runs before we start a new deploy.

At least, based on my reading of the github action docs, this should do what we want!

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - main
+
+concurrency:
+  group: "deploy"
+
 jobs:
   deploy:
     name: Deploy ngrok docs


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - main
+
+concurrency:
+  group: "deploy"
+  cancel-in-progress: true
```
jtszalay, line 10: Do we want to cancel the in progress? Or is it better to let it complete and the next to queue up?

jtszalay (approved):
Left a question, approved either way because concurrency limit is good to have here

.github/workflows/deploy.yml
euank: You know, you're probably right, it's probably better to leave the default pending behavior! I'll swap to that, thanks!

PR #1076, opened 2024-12-11
fix(overview-page): Wrapping of cards

jonstacks:
I think we are missing the flex classes on these divs. It leads to the docs overview wrapping weird:

image


jonstacks:
With the fix 
image

nijikokun (approved):

PR #1082, opened 2024-12-12
Add Traffic Policy Getting Started Guides for Agent Endpoints and Cloud Endpoints

nijikokun:
This creates traffic policy getting started guides for both agent endpoints and cloud endpoints:

Screenshot 2024-12-11 at 4 28 51 PM

It also fixes the `fromExact` redirect method by using `pathname` instead of `href` 👍


russorat (approved):
tinsel!

PR #1091, opened 2024-12-16
doc: correct typo in HTTP status from 201 to 401 Unauthorized

surrealist:
doc: correct typo in HTTP status from 201 to 401 Unauthorized


russorat (approved):
thank you!

PR #1115, opened 2025-01-08
fix: adding more info to error 9040

russorat:


TheConcierge (approved):
Seems reasonable to me

PR #1131, opened 2025-01-14
Testing better client-side redirects

S3Prototype:
WIP: This PR is trying to use docusaurus's built in client redirects plugin to do client- side redirects rather than our existing redirect script


S3Prototype:
Can't get this to work, and I need to reprioritize my time. For more info on why I pursued this, see this linear issue:
- https://linear.app/ngrok/issue/DOC-33/redirects-from-search-dont-work

PR #1132, opened 2025-01-14
Make URLs in search results absolute, clean up docs links, fix redirect script

S3Prototype:
Currently, because our docs site [uses client-side routing](https://ngrok.slack.com/archives/C03LRGNSG6A/p1736881450186529?thread_ts=1736867844.224459&cid=C03LRGNSG6A), our redirect script is not run when users visit links that use a relative path. (This is not relevant to users who visit the docs from external links)
  - When you click on links to routes that need to be redirected, no redirection happens.
  - If there is no page at the route when redirection fails, you get a 404.

This affects all links in the docs, and all docs search result items. (This issue does not affect Google search results)

Example: On [this page](https://ngrok.com/docs/traffic-policy/getting-started/agent-endpoints/cli/#step-3-test-it-out), select the link in the second bullet point that says `Check out the examples, use-cases and guides`. You get a 404 because the redirect fails.

### Temporary solution

This PR:

1. Removes stale links from the docs content. Instead of linking to old pages that redirect, we now link everywhere to the destination we wanted to redirect them to.
    - Test it:
      - In prod, visit [the godaddy guide](https://ngrok.com/docs/using-ngrok-with/godaddy/) and click the link text in the second sentence `added a reserved domain`. It should 404.
      - In the preview, [visit the same guide](https://ngrok-docs-git-shaquil-doc-36-swizzle-the-sear-722461-ngrok-dev.vercel.app/docs/using-ngrok-with/godaddy/) and click the same link. It will 404, then redirect.
    - You can do this for multiple other instances
1. Removes random links that are absolute URLs in the docs content for no reason. Maybe they were added by folks who didn't realize relative paths work.
1. Wraps results in the search modal with an `<a>` tag instead of `<Link>`. We provide the `<a>` with an absolute URL to cause a hard page load, triggering the redirect script.
1. Makes our redirect script account for trailing slashes
  - In some cases, [if we added a redirect with a trailing slash, it would not work without the slash](https://ngrok.slack.com/archives/C03LRGNSG6A/p1736871485920399?thread_ts=1736867844.224459&cid=C03LRGNSG6A). Not sure why, so I just updated the `fromIncludes` method to normalize the redirect it receives.
    - Test it:
      -  In prod, visit [https://ngrok.com/docs/traffic-policy/gallery](https://ngrok.com/docs/traffic-policy/gallery) (no trailing slash, so it 404s)
      - In the preview, [visit the same route](https://ngrok-docs-git-shaquil-doc-36-swizzle-the-sear-722461-ngrok-dev.vercel.app/docs/traffic-policy/gallery)
        - It 404s then redirects because the redirect is client-side. A server side redirect would prevent this 👍 

### Permanent solution (future)

To avoid this in the future, we should implement server-side redirects. Some googling seems to imply that this is possible with SPAs. If it's not realistic, we'd have to migrate to a meta framework like Next.js or Remix to leverage server side redirects.

Slack thread:
- https://ngrok.slack.com/archives/C03LRGNSG6A/p1736867844224459


nijikokun (approved):
This looks good, let's make sure we update the auto-generated docs in the other repository to avoid the hard-work here being undone.

PR #1139, opened 2025-01-17
Adding password reset link to error page

sachin20-ngrok:
Guidance for this error page was vague so I added the password reset link.


S3Prototype (approved):
Thanks!

PR #1145, opened 2025-01-21
remove extraneous examples folder

masonj5n:
For a few months our example generation pipeline was putting generated examples in this removed folder, which isn't actually referenced. The pipeline should now be correctly putting examples in `docs/api/resources/examples` so this folder shouldn't be used or written to anymore.


Megalonia (approved):

cody-dot-js (approved):
🦐

tammybailey (approved):

PR #1006, opened 2024-10-24
Update docs with new product architecture

nijikokun:
This adjust a lot of the integration and guides to use the new product architecture names.

PR #1105, opened 2024-12-31
rework documentation and documentation struture to reflect new product structure

inconshreveable:
This PR:

- Restructures the IA with a product-driven approach
- Deletes a bunch of outdated content
- Moves a ton of content to new areas
  - In the interest of avoiding redirects, not all content has been moved to new directories. Instead, they've been moved in the nav.
- Updates docs content to better reflect how we want to frame the product

### Requirements before merging

- [x] Double-check all redirects
- [x] Finish up the sections of the docs that say "TODO":
  - [x] Cloud Endpoints
  - [X] Public Endpoints
  - [x] K8s Architecture
  - [x] Agent Endpoints
  - [x] HTTP
- [x] /docs/universal-gateway/http/#url
  - [x] URLs section needs to be rethought
- [x] Merge main
- [x] Merge [Traffic Policy restructure docs branch](https://github.com/ngrok/ngrok-docs/pull/1173)

TODO

## Old description

This is a draft work in progress. Remaining items:

- [x] Agent tls termination guide is no longer linked from anywhere, find a new home for it.
- [x] Update links. A lot of pages got moved around, update all of the links to point to their new locations now.
- [x] Add redirects. A lot of pages got moved around, update all of the links to point to their new locations now.
- [ ] Implement TODOs
  - [ ] Cloud Endpoints
  - [ ] Public Endpoints
  - [ ] Agent Endpoints
  - [ ] HTTP
- [x] There are two Endpoints pages (Endpoints -> Overview and Concepts -> Endpoints) and they're currently duplicated which doesn't make sense. Figure something better out.
- [x] Endpoints Page needs to be written, most sections are missing.
- [x] Protocols > HTTP
  - [x] URLs section needs to be rethought
- [x] Merge main
-

PR #1176, opened 2025-02-14
Standardize CLI flag usage ('=' vs ' ')

joelhans:
I started to poke around at this, and think it's likely not as pervasive an issue as I had previously thought. 

I don't want to touch anything under the current **Endpoints** heading (HTTP, TCP, TLS), since those going away with #1006. I'm finding others here and there, but mostly in relation to modules, which, again, we are trying to move from anyway.

Anyway—I've found some and tested them to make sure they still work. I may come back for more later. Useful command to find instances, by replacing `url=` with the flag you're looking for.

```
grep -Ril --exclude-dir node_modules/ --exclude-dir build/ -e "url="
```


S3Prototype (approved):

S3Prototype:
Thanks for taking a look at this @joelhans! Yeah, no worries on stopping the work here. We'll get these changes merged into the revamp PR (#1105) and I'll make a linear issue to follow up and finish this off when that's done

PR #1181, opened 2025-02-20
Update basic-auth.mdx

bikerben33:
Removed a repetitive and confusing sentence. 
Fixed some typos.
Changed "more than 128 characters" to "less than 128 characters". I don't know the actual parameter, but logically it made sense that the password should be 8 to 128 characters.


S3Prototype:
So it looks like these changes are already implemented in the big IA restructure PR we landed. [You can see the updated page in prod](https://ngrok.com/docs/traffic-policy/actions/basic-auth/). Super sorry about that! We really appreciate you reaching out and making a PR.

PR #1182, opened 2025-02-20
Update index.mdx

bikerben33:
Fixed a few typos.
Reworded two sentences for better readability.
I prefer the oxford comma (the extra comma at the end of a set of three or more items) because in legal language this can get confusing. If I say split the proceeds evenly between A, B and C. One way to read that is that A gets 50% and the duo of B and C get the other 50%. In technical terms you might specify to divide resources between A, B and C and a user might not know if they should give A 50% or 33.33%.


S3Prototype:
Yeah, this page is deleted as well. Apologies

PR #1183, opened 2025-02-20
Update index.mdx

bikerben33:
I believe SNI is Server Name Indication so I changed that reference. I'm not sure why Domain is capitalized some times and sometimes not. In my opinion, it should not be capitalized unless it's a proper name (like Domain CityGate Luxury Apartments). Found a few typos.


S3Prototype:
This has also been fixed already

PR #1187, opened 2025-02-20
Update Coming Soon section in openid-connect.mdx

bikerben33:
The use of the word "interim" is harder to translate for English as a Second Language (ESL) readers. Also, it was confusing to say OpenID works similarly but not reference what it is similar to.


S3Prototype:
Yeah, this has also already been addressed unfortunately, but thanks for making the PR

PR #1196, opened 2025-03-04
Add tutorial: Deploy a multicloud API gateway

joelhans:
This supplants #1161, which I'll close momentarily.

After talking to @shubcodes, we agreed that endpoint pools are not the right strategy for the most common multicloud requirements. Endpoint pools could become an *advanced* section of this tutorial, or a follow-up tutorial.

Also not included here: failover strategy using internal cloud endpoints. I think the right way is to ship this and write up failover as a follow-on task later this month.


shubcodes:
Couple of comments before merge: 

1. I do not like the >1 in APIs and clouds, just say multiple or more than one
2. the service names with multiple 'dashes' like 'abc-app-service' gets super confusing with the internal endpoint names like "https://abc-cloud-a.internal" 
3. show creating a cloud endpoint with the api as well
4. i think it might be important to explain traffic policy phases
5. i need the so what to be a little more explained; so what we can do JWT validation? maybe if we show JWT we should also show OAuth for browser
6. lastly; give me some diagrams i want to look at something -- reference the diagrams I used in the demo slides

joelhans:
@shubcodes Thank you! Responding to them here.

1. I do not like the >1 in APIs and clouds, just say multiple or more than one

Fixed.

3. the service names with multiple 'dashes' like 'abc-app-service' gets super confusing with the internal endpoint names like "https://abc-cloud-a.internal" 

Fixed. Apps are now `app` and services `abc-service`, which simplifies things _a bit_.

5. show creating a cloud endpoint with the api as well

Done. Did this for both creating the cloud endpoint and updating it with new Traffic Policy.

7. i think it might be important to explain traffic policy phases

Done. Explained phases and expressions more.

9. i need the so what to be a little more explained; so what we can do JWT validation? maybe if we show JWT we should also show OAuth for browser

Added some _why_ for JWTs. I'm going to push back a little on OAuth and argue that it's out of scope for this—I want to be really focused on APIs, not necessarily a mix of APIs and web services, and I think adding OAuth in really throws a wrench in an otherwise mostly streamlined setup. Would I need to make one of the two services a webapp to clarify? I know that's useful for demoing ngrok as a universal gateway, but this already feels quite hefty.

11. lastly; give me some diagrams i want to look at something -- reference the diagrams I used in the demo slides

Made a quick diagram for the final architecture. I don't love the idea of having diagrams inserted throughout, but I could be convinced to do that as a follow-on improvement when we can grab some time from Micah to whip them up into a consistent style.

PR #1201, opened 2025-03-04
Streamline+update Auth0 integration guide

joelhans:
Suggesting some improvements to this guide based on my own experiences setting up JWT for #1196, and because this referenced edges and the old Traffic Policy editor.

I'd like to have a solid and up-to-date guide for reference whenever we're talking about setting up JWTs with Auth0 instead of repeating the content.


S3Prototype (approved):
Thanks!

PR #1202, opened 2025-03-04
Update format of variables used in digitalocean gslb integration

rkolavo:


joelhans (approved):
Oh, this guide! I'd forgotten all about it. Thank you for the fix—I should update this whole thing to use endpoint pools.

PR #1209, opened 2025-03-06
Change Custom Response status_code to optional

kate-osborn:
The `status_code` field for the Custom Response action is now optional and has a default value of `200`. 

Also, this PR fixes an issue with the Custom Response page where sections were duplicated and in the wrong layout.

Previous Custom Response TOC:
Screenshot 2025-03-06 at 1 04 07 PM

New TOC:
Screenshot 2025-03-06 at 1 04 40 PM


S3Prototype (commented):
Thanks

S3Prototype (approved):

PR #1211, opened 2025-03-06
Potentially fix integrations pages

joelhans:
I noticed earlier today that all the integrations pages are broken—the [integrations hub](https://ngrok.com/docs/integrations/) is empty, as are the individual hubs for each provider.

Error I'm seeing on any integrations page:

```
ZodError: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      11,
      "docs",
      0,
      "contentTitle"
    ],
    "message": "Required"
  }
]
```

I poked around a bit and figured that error was getting thrown because not every integrations guide explicitly has a `title:` in the frontmatter, and this Zod dude doesn't like an undefined required string. Making `contentTitle` optional _seems_ to fix the error and restore the integrations hubs... but at what cost?

### DIFF
--- a/src/components/IntegrationPageList.tsx
+++ b/src/components/IntegrationPageList.tsx
@@ -22,7 +22,7 @@ export default function IntegrationPageList({ name }: Props) {
 						title={
 							doc.frontMatter?.title ||
 							doc.contentTitle ||
-							doc.frontmatter?.name ||
+							doc.frontMatter?.name ||
 							"Integration"
 						}
 						description={doc.frontMatter?.description || doc.excerpt}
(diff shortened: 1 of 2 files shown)


S3Prototype:
@joelhans I fixed this in a separate PR that's not gonna merge soon, so I added those changes here. Let's see if it works after CI runs

S3Prototype (approved):

src/components/IntegrationPageList.tsx
```diff
@@ -19,7 +19,12 @@ export default function IntegrationPageList({ name }: Props) {
 					<NgrokCard
 						to={doc.path}
 						size="sm"
-						title={doc.frontMatter?.title || doc.contentTitle}
+						title={
+							doc.frontMatter?.title ||
+							doc.contentTitle ||
+							doc.frontMatter?.name ||
```
S3Prototype, line 25:
```suggestion
```

joelhans:
@S3Prototype Amazing, thanks for carrying that over the finish line!

PR #1217, opened 2025-03-11
Add agent config snippets to all config examples

S3Prototype:
This PR:
- Updates the `ConfigExample` component to show an accordion alongside traffic policy snippets. The accordion contains an example snippet with the same traffic policy in a full agent config file.
  - [See it here](https://ngrok-docs-git-shaquil-doc-77-improve-indentat-5afed3-ngrok-dev.vercel.app/docs/traffic-policy/examples/block-unwanted-requests/#deny-traffic-from-tor-networks) 
  image
- Adds a `clean-dev` command to `package.json`, because sometimes docusaurus errors out on build and requires npm cache clean to work again


src/components/ConfigExample.tsx
```diff
@@ -94,61 +102,66 @@ export type ConfigExampleProps = {
...
-					{createElement(components.h3, { id: "policy" }, "Policy")}
-				</>
+				<Accordion type="multiple" defaultValue={["show-agent-config"]}>
+					<AccordionItem value="show-agent-config1">
+						<AccordionHeading asChild>
+							<p className="text-sm font-medium text-gray-600">
+								{!configShowing
+									? "Show agent config example"
+									: "Hide agent config example"}
+								<AccordionTrigger
```
jhou, line 149: just curious (as I know little about our docs system yet), this means that one can only click on the triangle to expand the Agent Config example? 

If so, does it still work if the "Show/Hide....." text to be included?

S3Prototype: @jhou Yeah, only the arrow will expand/collapse it, but we can make the text clickable as well by moving it into the `<AccordionTrigger>`

nijikokun (approved):
Great change @S3Prototype! We will likely want to switch the implementation when we have code blocks with tabs, but this looks good for now.

PR #1220, opened 2025-03-11
Restructure Action Result Variable docs

S3Prototype:
wip


S3Prototype:
After discussion with Niji, we're gonna put this on hold and work on restructuring the Universal Gateway docs first

PR #1229, opened 2025-03-14
Mark JWT aud claim as optional

iambmelt:
Per [RFC-7519](https://datatracker.ietf.org/doc/html/rfc7519), use of the `aud` claim is [optional](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3:~:text=generally%20application%20specific.-,Use%20of%20this%20claim%20is%20OPTIONAL,-.%0A%0A4.1.4.%20%20%22exp). Marking this field as `required=false` accordingly.


iambmelt:
Converting to draft, based on commentary from Ryan that this property is currently required by the backend.

iambmelt:
Diff showing zero lines changed... 🧐

andrew-harris-at-ngrok:
> Diff showing zero lines changed... 🧐

This appears to have been resolved in #1245 -- https://github.com/ngrok/ngrok-docs/commit/31e67230109d8361789408cb1cb4c165c1ad48bd

I would recommend closing this MR without merge.

PR #1233, opened 2025-03-18
Adding an Example For Wildcard Endpoints 

sachin20-ngrok:
[https://ngrok.slack.com/archives/C053S5EUL69/p1741211127345889]

Convo on this section for context ^^^

Adding a section in the universal gateway portion of docs that demonstrate to users that they can start wildcard endpoints on a given subdomain from anywhere without having to reserve subdomains


S3Prototype (approved):
Thanks @sachin20-ngrok!

In an interlock, we decided to drop using the `=` in Agent CLI code snippets e.g. `ngrok http 80 --domain = https://whatever.com` where unnecessary. I removed the `=` from both snippets. Will they still work? If so, I think you can go ahead and merge this 👍 

I also made some changes:

I simplified the language and brought the examples directly into the page since they're so simple. IMO we should only import examples when they're going to be reused in multiple places or they're so big they make the page hard to edit.

Sidenote: The `<Tabs>` component is only necessary if you need to switch between more than one option, so no need for it in this case.

docs/universal-gateway/http.mdx
```diff
@@ -369,6 +369,24 @@ will receive traffic for `https://foo.example.com` and
...
+
+```bash
+ngrok http 80 --domain https://api1.example.com
+```
+
+While the second is at port `81`:
+
+```bash
+ngrok http 81 --domain https://api2.example.com
+```
```
S3Prototype, line 388:
````suggestion
```bash
ngrok http 80 --url https://api1.example.com
```

While the second is at port `81`:

```bash
ngrok http 81 --url https://api2.example.com
```
````

File: docs/universal-gateway/http.mdx [hunk 2]
```diff
@@ -369,6 +369,24 @@ will receive traffic for `https://foo.example.com` and
 	</TabItem>
 </Tabs>
 
+### Use multiple endpoints with wildcard subdomains
+
+With the [Agent CLI](/agent/cli), you can create separate endpoints for subdomains of a wildcard domain no matter where the upstream services are running.
+
+For example, if you reserve `*.example.com` and want to route requests to `api1.example.com` and `api2.example.com`, you can specify subdomains via the CLI and ngrok will handle routing requests to the correct endpoint, even if they're on different ports.
```
S3Prototype, line 376:
```suggestion
For example, if you reserve `*.example.com` and want to route requests to `api1.example.com` and `api2.example.com`, you can specify subdomains via the CLI. ngrok will handle routing requests to the correct endpoint, even if they're on different ports.
```

docs/universal-gateway/http.mdx
```diff
@@ -369,6 +369,24 @@ will receive traffic for `https://foo.example.com` and
 	</TabItem>
 </Tabs>
 
+### Use multiple endpoints with wildcard subdomains
+
+With the [Agent CLI](/agent/cli), you can create separate endpoints for subdomains of a wildcard domain no matter where the upstream services are running.
```
S3Prototype, line 374:
```suggestion
With the [Agent CLI](/agent/cli), you can create separate public endpoints for subdomains of a wildcard domain no matter where the upstream services are running. See [the Domains docs](/docs/universal-gateway/domains/#wildcard-domains) for more information.
```

docs/universal-gateway/http.mdx
```diff
@@ -369,6 +369,24 @@ will receive traffic for `https://foo.example.com` and
...
+```bash
+ngrok http 80 --url https://api1.example.com
+```
+
+While the second is at port `81`:
+
+```bash
+ngrok http 81 --url https://api2.example.com
+```
+
```
S3Prototype, line 389:
```suggestion

:::tip
You can only create public endpoints for subdomains of wildcard domains.
:::
```

docs/universal-gateway/http.mdx
```diff
@@ -369,6 +369,28 @@ will receive traffic for `https://foo.example.com` and
...
+:::tip
+You can only create public endpoints for subdomains of wildcard domains.
```
S3Prototype, line 391:
```suggestion
This is only necessary for creating _public_ endpoints to subdomains. Internal endpoints don't require a domain.
```

sachin20-ngrok:
@S3Prototype urls work without the "=" since, should be good to merge!

PR #1249, opened 2025-03-26
Update lang switcher

S3Prototype:
The current lang switcher wraps codeblocks in tabs. We actually want to embed the tabs in the codeblock header.

Now the switchers are in the header, are saved to localstorage, and update all codeblocks at once:

https://github.com/user-attachments/assets/6e357359-1db3-4ffd-8484-fed1fc8d8647

Which is really useful with our ConfigExample component. This allows us to show both Agent Config and Traffic Policy examples with one set of tabs, and you can swap between languages within the codeblock.

https://github.com/user-attachments/assets/f79c4afd-dfe3-4888-b456-739cdf8d9583

Our current solution in prod is to tuck the agent config example in a dropdown, which is less optimal:

https://github.com/user-attachments/assets/d61fe1bd-3063-4d6d-821c-83306d5ff77a

## Testing

1. You can visit a page with language tabs, like [one of the traffic policy examples](https://ngrok-docs-git-shaquil-doc-226-update-lang-swi-0e3c30-ngrok-dev.vercel.app/docs/traffic-policy/examples/enforce-tls/). YAML should be the default language tab.
2. Then select JSON in the language switcher
3. Then [go to a different page with a language switcher](https://ngrok-docs-git-shaquil-doc-226-update-lang-swi-0e3c30-ngrok-dev.vercel.app/docs/traffic-policy/examples/compress-json-responses/), and JSON should be the default instead of YAML
4. On a page with multiple examples [like this](https://ngrok-docs-git-shaquil-doc-226-update-lang-swi-0e3c30-ngrok-dev.vercel.app/docs/traffic-policy/examples/a-b-tests/#basic-example), select a language and watch both examples switch language


joelhans (approved):
This is amazing—makes the UX around these snippets and the YAML/JSON/file/agent Traffic Policy ordeal way more cohesive.

PR #1261, opened 2025-03-31
Explain how to run documentation in one line, using Docker

RichardJECooke:
Add paragraph to README explaining how to these docs with a single command. Using Docker also isolates developers' machines from any malicious JS libraries and prevents any clutter on your host environment.


S3Prototype:
Thank you!

PR #1265, opened 2025-03-31
Adding a note on the wildcard domains page to include how subdomains …

sachin20-ngrok:
https://linear.app/ngrok/issue/DOC-64/update-domain-pricing-to-include-information-about-subdomains


S3Prototype (approved):
Thanks!

PR #1266, opened 2025-03-31
Adds example of custom-response for internal endpoint downtime

kaganjd:
Preview: https://ngrok-docs-git-jen-add-example-ngrok-dev.vercel.app/docs/traffic-policy/actions/custom-response/#custom-response-for-internal-endpoint-downtime


kaganjd:
@S3Prototype do you know how I can remove the agent endpoint switcher for this example? It only works on a cloud endpoint, so would like to remove the agent endpoint text for clarity

PR #1274, opened 2025-04-07
update dependencies; remove need to convert tabs to spaces and that dep

cody-dot-js:
1. update dependencies + pnpm version
2. updates mantle: fixes an issue where the initial render of the code block would use unescaped html as dangerouslySetInnerHtml
   1. all rendered lines now use spaces as indentation by default, no need to convert tabs

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ngrok-doc",
   "version": "0.0.0",
   "private": true,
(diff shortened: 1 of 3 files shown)


nijikokun (approved):
Checked out the code snippets looks good :) We can ignore copilot here I think.

PR #1277, opened 2025-04-09
swap prettier out for biome (gotta go fast)

cody-dot-js:
prettier is really, really slow

biome lints AND formats the whole repo in ~200ms

### DIFF
--- a/src/components/LangSwitcher/index.tsx
+++ b/src/components/LangSwitcher/index.tsx
@@ -18,10 +18,11 @@ import LangSwitcherContext, {
 } from "./LangSwitcherContext";
 import { getCodeBlocks } from "./utils";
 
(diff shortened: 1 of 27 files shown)


src/components/LangSwitcher/index.tsx
```diff
@@ -18,30 +18,35 @@ import LangSwitcherContext, {
 } from "./LangSwitcherContext";
 import { getCodeBlocks } from "./utils";
 
+// biome-ignore lint/suspicious/noExplicitAny: <explanation>
```
cody-dot-js, line 21: theres's a lot of rough edges in this file (and the sibling `utils` file), i opted to simply fix the typescript errors and `// biome-ignore` the rest (mostly `:any`)

nijikokun (approved):
LGTM

PR #1307, opened 2025-04-26
add validate-codeblocks CI linter

inconshreveable:
There are a lot of invalid code blocks in our documentation.

1. Write a script which scans all of the codeblocks in our docs repo and fails if any of them do not pass a syntax check.
2. Fix all of the problems that the script caught. There are many. Lots of our k8s yaml manifests had tabs (which are invalid in YAML), lots of improper indentation, etc.

The actual script of interest is in src/validate-codeblocks/validator.ts

- This script currently only works with JSON/YAML codeblocks. Other lanauges will need support added for them in the future.
- There are a few codeblocks that we want syntax highlighting on even though they are intentionally wrong in a few places, for example where we have ellipses to indicate elided output. In these cases you can now annotate those code blocks with a `skip-validation` tag and the codeblock validator will ignore them.


jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
cody-dot-js, line 1: Why jest? Should prefer vitest. It's faster and less quirky, supports ts and esm out of the box

package.json
```diff
@@ -21,7 +21,12 @@
     "serve": "docusaurus serve",
     "typecheck": "tsc",
     "write-translations": "docusaurus write-translations",
-    "write-heading-ids": "docusaurus write-heading-ids"
+    "write-heading-ids": "docusaurus write-heading-ids",
+    "build:validate-codeblocks": "tsc -p src/validate-codeblocks/tsconfig.json",
+    "validate-codeblocks": "pnpm run build:validate-codeblocks && node dist/validate-codeblocks/index.js",
```
cody-dot-js, line 26: If we use tsx, you can run typescript files directly without having to precompile them

src/validate-codeblocks/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
cody-dot-js, line 1: We don't need this, should be able to use the root tsconfig

src/validate-codeblocks/validator.ts
```diff
@@ -0,0 +1,272 @@
+import fs from "node:fs/promises";
+import chalk from "chalk";
+import globby from "globby";
```
cody-dot-js, line 3: globby bad. tinyglobby good

src/validate-codeblocks/validator.ts
```diff
@@ -0,0 +1,272 @@
+import fs from "node:fs/promises";
+import chalk from "chalk";
```
cody-dot-js, line 2: node 22 has native text styling now btw

https://nodejs.org/docs/latest-v22.x/api/util.html#utilstyletextformat-text-options

tsconfig.json
```diff
@@ -37,7 +37,8 @@
     "types": [
       "node",
       "@docusaurus/module-type-aliases",
-      "@docusaurus/theme-classic"
+      "@docusaurus/theme-classic",
+      "jest"
```
cody-dot-js, line 41: we should remove this and replace jest with vitest

inconshreveable:
all issues are - i believe - fixed

S3Prototype (approved):
LGTM

PR #1311, opened 2025-04-29
docs: add auth to your endpoint using policy

jnbritton:
#CON-145

This is a step-by-step guide for how users can get started applying authentication to their endpoints using Traffic Policy from their CLI environment. 

There are specific steps and callouts here for cases when (a) users are using a separate traffic policy YAML file vs (b) using an ngrok.yml config file. If the product/content teams decide we should add more nuanced steps for how to set this up in both paths (a) and (b). I'd also be open to adding more content on how to set this up (c) if users are using the ngrok Dashboard to set this up.


S3Prototype (approved):
Thanks for this! I've added my changes. Let me know if I cut out some important context/info, and feel free to add anything you feel is still missing 👍

PR #1319, opened 2025-05-01
Update titles and sidebar labels to differentiate page titles for Overviews

joelhans:
I noticed this a long time ago, forgot about it, but then noticed it again today—the page titles for our **Overview** docs are all **Overview | ngrok Documentation**, which makes differentiating between them hard if you have multiple tabs open.

I decided to go ahead with this solution, but also open to suggestions on where/how we should specify on-page vs. sidebar vs. SEO titles. I think there's at least one other way we might do this.

Before:

Screenshot 2025-05-01 at 11 21 35 AM

After:

Screenshot 2025-05-01 at 11 39 01 AM

Good news, and potential side quest opportunity—looks like Google ignores page titles on search results, so this hasn't been an issue when people are searching for our docs on Google. But how/why does Google just ignore our titles and make its own?

image
image

### DIFF
--- a/docs/agent-sdks/index.mdx
+++ b/docs/agent-sdks/index.mdx
@@ -1,6 +1,7 @@
 ---
 sidebar_position: 1
-title: Overview
+title: Agent SDKs
+sidebar_label: Overview
 pagination_next: k8s/index
 ---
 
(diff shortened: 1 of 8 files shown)


S3Prototype (approved):

S3Prototype:
Thanks!

nijikokun:
> But how/why does Google just ignore our titles and make its own?
AI

PR #1330, opened 2025-05-05
Add new guide for load balancing between clouds

joelhans:
[#CON-131](https://linear.app/ngrok/issue/CON-131/how-do-i-load-balance-across-clouds)

One last quick guide to push through for using endpoint pools across clouds. I decided to also bring in the environment agnostic part of this with the Go SDK, but it definitely adds a tiny bit of bloat—open to feedback on that part. This then points to using a Cloud Endpoint + Traffic Policy, which sets them up better to use the mentioned custom strategies.

Once this and #1322 are through, I'm going to create a new PR that removes old load balancing guides via Edges and pushes them to one of these three guides.

### DIFF
--- a/docs/agent-sdks/index.mdx
+++ b/docs/agent-sdks/index.mdx
@@ -1,6 +1,7 @@
 ---
 sidebar_position: 1
-title: Overview
+title: Agent SDKs
+sidebar_label: Overview
 pagination_next: k8s/index
 ---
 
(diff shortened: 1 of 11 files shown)


joelhans:
@S3Prototype Thanks! I like the structure you came up with. I wanted to show two things here, 1) the multicloud angle (obviously), and 2) that pooling works across tools/form factors, so this feels like a good balance between those goals.

I made a few small tweaks to language to finish the "conversion," so feel free to give it one more pass for style.

S3Prototype (approved):

PR #1338, opened 2025-05-07
Remove old load balancing guides (Edges) and redirect to new one

joelhans:
In working on the latest few load balancing guides, I saw that we still have some that reference Edges. We definitely shouldn't have both, as that would cause confusion, and we want people to use pools, not tunnel backends or anything like that.

Redirecting all of these to the multiple cloud guide might not be the best long-term, but I think it's the most appropriate of the three new guides, given that it covers multiple tools (agent, CLI) instead of *just* focusing on K8s like my other two new guides.


S3Prototype (approved):
TY

PR #1353, opened 2025-05-15
Update custom domains guide

S3Prototype:
- Moves it to a better location in the sidebar
- Updates guide instructions
- Renames kubernetes custom domain guide so it stops showing up at the top of search results

### DIFF
--- a/docs/k8s/guides/custom-domain.mdx
+++ b/docs/k8s/guides/custom-domain.mdx
@@ -1,5 +1,6 @@
 ---
-title: Custom Domains
+sidebar_label: Custom Domains
+title: Kubernetes Custom White Label Domains
 ---
 
 In the Kubernetes Ingress and Gateway API specs, ingresses and routes can have multiple rules with different hostnames. While standard ngrok domains are available for use immediately after reservation, custom white label domains may require a couple extra steps to get working. The following outlines 2 options for getting custom white label domains working with the ngrok Kubernetes Operator.
--- a/docs/universal-gateway/custom-domains.mdx
+++ b/docs/universal-gateway/custom-domains.mdx
@@ -46,13 +46,13 @@ To start an agent endpoint at your domain, run the following command in your ter
 
 ## Wildcard domains
 
-If you bring your own [wildcard domain](/universal-gateway/wildcard-domains), e.g. `*.example.com`, you will need to create a second `CNAME` record with your domain host for [wildcard TLS Certificate provisioning](/universal-gateway/tls-certificates/how-does-ngrok-handle-tls/#wildcard-domains).
+If you bring your own [wildcard domain](/universal-gateway/domains#wildcard-domains), e.g. `*.example.com`, you will need to create a second `CNAME` record with your domain host for [wildcard TLS Certificate provisioning](/universal-gateway/tls-certificates/#wildcard-domains).
 
 ## Apex domains
 
 Because it's not possible to create `CNAME` records for apex domains, if you want to use an apex domain, e.g. `example.com`, you must use a DNS provider that supports an `ALIAS` record or `CNAME` flattening.
 
-Because of how `ALIAS`/`CNAME` flattening is implemented, apex domains will not take advantage of the [Global Load Balancer](/universal-gateway/global-load-balancer/#domains). If you're trying to create your apex domain because you need to create endpoints for multiple subdomains, use a [wildcard domain](/universal-gateway/what-are-public-endpoints#wildcard-domains) instead.
+Because of how `ALIAS`/`CNAME` flattening is implemented, apex domains will not take advantage of the [Global Load Balancer](/universal-gateway/global-load-balancer/). If you're trying to create your apex domain because you need to create endpoints for multiple subdomains, use a [wildcard domain](#wildcard-domains) instead.
 
 ## Using custom domains with TCP endpoints
 
--- a/docs/universal-gateway/domains.mdx
+++ b/docs/universal-gateway/domains.mdx
@@ -86,30 +86,8 @@ For example, if you create a Domain with the name `foo.ngrok.app`:
 
 ## Bring your own domain {#branded-domains}
 
-You can use any domain name that you already own with ngrok, e.g.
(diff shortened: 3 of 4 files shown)


docs/universal-gateway/custom-domains.mdx
```diff
@@ -0,0 +1,70 @@
...
+---
+
+This guide will show you the steps to use any domain name that you already own with ngrok, e.g. `app.your-domain.com`.
+
+:::note
+ngrok is not a domain registrar; you must already own a domain name to use it
+with ngrok.
+:::
+
+## 1. Create a domain in ngrok
```
sg-writer, line 14: maybe this is beyond the scope of this PR—I don't know if this terminology is used widely—but to "create" a domain in ngrok feels like the wrong word choice to me. Even after the warning that ngrok is not a registrar, "create" still makes me second-guess whether the callout is true. Maybe we can replace it with "add"? ("your" could also help clarify that you should already have it)

```suggestion
## 1. Add your domain in ngrok
```

docs/universal-gateway/custom-domains.mdx
```diff
@@ -0,0 +1,70 @@
...
+When your domain is created, you'll be given a `CNAME` value such as `exampledata.otherdata.ngrok-cname.com`. Copy it.
+
+:::info
+You can also create a Domain via the [ngrok API](/api/resources/reserved-domains/#request). The `CNAME` will be in the `"cname_target"` field of [the response object](/api/resources/reserved-domains/#response).
+:::
+
+## 2. Add the CNAME record to your domain
+
+1. In a new browser tab, navigate to your domain hosting provider's dashboard and create a CNAME record. Check your hosting provider's support content or documentation to learn how.
+   - Typically, this process includes creating a [DNS record](https://en.wikipedia.org/wiki/List_of_DNS_record_types).
```
sg-writer, line 27: I would cut this line—I don't think it's useful to send the reader to this Wikipedia page while they're in the middle of this setup.

```suggestion
```

docs/universal-gateway/custom-domains.mdx
```diff
@@ -0,0 +1,70 @@
...
+:::tip Test your CNAME record
+Using a tool like [`dig`](https://www.digwebinterface.com/) or [`nslookup`](https://www.nslookup.io/), you can test your CNAME record in the terminal.
+:::
+
+## 3. Start an endpoint using your domain:
+
+To start an agent endpoint at your domain, run the following command in your terminal, replacing `your-domain-here.example.com` with the domain you created in the ngrok dashboard.
+`bash
+	ngrok http --url your-domain-here.example.com 8080
+	`
```
sg-writer, line 42: ` ``` `

docs/universal-gateway/custom-domains.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+## Apex domains
+
+Because it's not possible to create `CNAME` records for apex domains, if you want to use an apex domain, e.g. `example.com`, you must use a DNS provider that supports an `ALIAS` record or `CNAME` flattening.
+
+Because of how `ALIAS`/`CNAME` flattening is implemented, apex domains will not take advantage of the [Global Load Balancer](/universal-gateway/global-load-balancer/). If you're trying to create your apex domain because you need to create endpoints for multiple subdomains, use a [wildcard domain](#wildcard-domains) instead.
+
+## Using custom domains with TCP endpoints
+
+Public TCP endpoints are assigned randomly on an ngrok-controlled hostname with a randomly-assigned port. You may not choose the hostname and you may not select the port.
```
sg-writer, line 56: - no need to hyphenate adverbs
- preference for "cannot" over "may not"

```suggestion
Public TCP endpoints are assigned randomly on an ngrok-controlled hostname with a randomly assigned port. 
You cannot choose the hostname or select the port.
```

sg-writer (commented):
bunch of editorial touchups and suggestions here—accept or reject at your discretion. If you like the style rules I've suggested in the `custom-domains` doc then I'm happy to add them to the style guide. The only thing really preventing me from approving are the typos (which I recognize were already there before)—I can't let "dowmain" through! 😁

docs/k8s/guides/custom-domain.mdx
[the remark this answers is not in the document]
S3Prototype: @sg-writer  I don't personally think this one would be worth all the redirects and link updates we would have to do. Because this project is in DocuSaurus, navigation is done in the client. This means the redirect script doesn't run when clicking links to docs pages from within the docs. As a result, we'd have to update all links across the docs, which is prone to error and not quite worth it imo

docs/universal-gateway/custom-domains.mdx
```diff
@@ -0,0 +1,70 @@
+---
+title: How Do I Use A Custom Domain?
+sidebar_label: Using Custom Domains
+description: Learn how to use your own custom domain with ngrok.
+---
+
+This guide will show you the steps to use any domain name that you already own with ngrok, e.g. `app.your-domain.com`.
```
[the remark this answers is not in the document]
S3Prototype: I like this, so let's add it to the style guide. But I think the e.g. battle is going to be hard to win with engineers. They seem to love that phrase, lol @sg-writer

docs/k8s/guides/custom-domain.mdx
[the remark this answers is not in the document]
sg-writer: Ah yeah that's frustrating, in that case I'm ok with leaving it alone!

docs/universal-gateway/custom-domains.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+:::tip Test your CNAME record
+Using a tool like [`dig`](https://www.digwebinterface.com/) or [`nslookup`](https://www.nslookup.io/), you can test your CNAME record in the terminal.
+:::
+
+## 3. Start an endpoint using your domain
+
+To start an agent endpoint at your domain, run the following command in your terminal, replacing `your-domain-here.example.com` with the domain you added in the ngrok dashboard:
+
+````bash
```
sg-writer, line 42:
```suggestion
```bash
```

PR #1354, opened 2025-05-16
small updates to owasp docs

benjaminchan:
- formatting: OWASP/CRS -> OWASP CRS
- document CRS version
- document behaviors that result in http 403 / 500
- process body default to false
- block and test mode examples
  - removed the basic examples which detail example traffic + matching rules until we standardize them further, they are helpful and we do want those


benjaminchan:
added the example req/res back in

S3Prototype (approved):

PR #1362, opened 2025-05-21
Add video to Endpoint Pooling doc

joelhans:
Pretty self-explanatory!

Not sure if this is the right implementation or the best spot, but it felt natural to me.


samcrichard (approved):

inconshreveable:
1. suggest we bring it up to the bottom of the intro section so that it's above the fold.
2. setting a fixed width breaks this on mobile/tablet, etc. think we need a different solution there

joelhans:
@cody-dot-js lol I was just thinking we should make a component to make this easier in the future. 🫶 Well done!

cody-dot-js:
fixed the build, renders well on mobile, and we have a reusable component to use now

image

PR #1364, opened 2025-05-22
Add new pooling video to K8s guides

joelhans:
I pushed out a new K8s-specific pooling video today, so using the new `YouTubeEmbed` component to drop it in a few places. Plus, two other pools-related tweaks:

- Fixed typos in the pooling doc
- Changed the link for **Load Balancer** in _What is ngrok?_ to point to pools.


S3Prototype:
I added some margin to the YouTube embed component, and if that looks good we can get this merged. Just waiting on the preview to build 👍

PR #1374, opened 2025-05-30
Whitelist the Algolia crawler in our robots.txt

S3Prototype:
We need to whitelist the Algolia crawler in our robots.txt so it will stop being rate-limited when it's indexing our site. When the crawler is rate-limited, some docs will not be included in the search index.

https://ngrok.slack.com/archives/C03LRGNSG6A/p1748560268249429


S3Prototype:
This was handled with a PR to the ngrok private repo.

PR #1378, opened 2025-06-02
Iterate on Universal Gateway Overview description

sg-writer:
Following internal discussions on Slack, I'm proposing an update to the introductory description of Universal Gateway on the Overview page. I believe this description is more accurate and easier for new users to understand.

While here, I also took a minute to enforce the one-sentence-per-line style rule and added an Oxford comma toward the bottom of the page.

### DIFF
--- a/docs/universal-gateway/overview.mdx
+++ b/docs/universal-gateway/overview.mdx
@@ -6,16 +6,13 @@ pagination_next: universal-gateway/domains
 
 # Universal Gateway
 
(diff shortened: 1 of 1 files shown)


S3Prototype (approved):

nijikokun:
Just going to call out that the second paragraph has different concepts than the ones listed under it. We should re-think this page, it becomes blurry what the purpose is after concepts.

nijikokun (approved):
Approved, but we should revisit this page conceptually.

PR #1381, opened 2025-06-03
Remove unnecessary index page for other guides directory

S3Prototype:
We have a top-level index page in the other guides directory that is an unmaintained list of all the other guides in that directory. This is unnecessary because it just recreates the sidebar, so we should delete it.


sg-writer:
Is this pattern an ngok thing or a docusaurus thing? I do wonder if there was a good reason for these to exist once upon a time. Maybe for SEO reasons (so that _something_ exists at these URLs)? Is there a softer solution that involves removing the links from the sidebar without deleting the pages?

sg-writer (approved):
All things considered I don't have a strong opinion about these pages and would prefer not to see them on the sidebar! I have been frequently confused by them as a new user!

S3Prototype:
@sg-writer yeah, these are good questions. In my experience, we've never needed a top-level index page in Docusaurus, and we have multiple sections that don't have top-level index pages already. A lot of these docs were written by engineers more than a year ago who weren't all that familiar with Docusaurus, so there's a lot of decisions that have been made for these docs that a docs team would not have made.

Abdiifahman (commented):
🤣

PR #1393, opened 2025-06-09
misc k8s docs improvements

Alice-Lilith:
Addresses several smaller friction and confusion points that were identified with the k8s operator docs


docs/getting-started/kubernetes/crds.mdx
```diff
@@ -48,7 +48,10 @@ Find more details on our [pricing page](https://ngrok.com/pricing), or, if you'r
 
 ## 4. Create your endpoint manifest
 
-You have two options to expose your Kubernetes service with ngrok's CRDs:
+You have two options to expose your Kubernetes service with ngrok's CRDs. The primary two types are `CloudEndpoint` and `AgentEndpoint`.
+When using other types of configuration with the operator such as `Ingress` and Gateway API resources, the operator translates all of them into
+`CloudEndpoint` and `AgentEndpoint` resources. Using the `CloudEndpoint`/`AgentEndpoint` resources directly gives you maximum control over your ngrok operator configuration.
+For more information about cloud and agent endpoints, see [the endpoint types page](/docs/universal-gateway/types/).
```
sg-writer: this addition would seem to flow more naturally if it came after the numbered list. So you get the breakdown of what these two options are first, and then additional clarifying details and a link to further information. Then the example manifest.

sachin20-ngrok (approved):

joelhans:
Having the quickstarts links in both places in the sidebar creates a kind of strange jumping behavior if I click between them. I feel like @S3Prototype had a solution for this in a previous iteration of the k8s docs that we could bring back here?

https://github.com/user-attachments/assets/426f9d3e-15dc-4161-9c2c-3b5f41c26838

Otherwise 👍 👍 👍

S3Prototype (dismissed):
I'll try to fix the sidebar issue with this and then we can merge it.

Alice-Lilith:
> Having the quickstarts links in both places in the sidebar creates a kind of strange jumping behavior if I click between them. I feel like @S3Prototype had a solution for this in a previous iteration of the k8s docs that we could bring back here?
> Screen.Recording.2025-06-10.at.8.00.48.AM.mov
> 
> Otherwise 👍 👍 👍

I feel like the quickstart shouldn't have been moved out of the relevant sections and it would be more appropriate for them to live in the directories for the relevant sections, but I get the need to surface all the quickstart at the top of the list as well so they're easier to find without digging for them.

IMO the quickstarts should live in their relevant sections, then if we want a "quickstart aggregator" section at the top, clicking on the relevant quickstart will jump you to the relevant section, which I think would be a better flow.

PR #1400, opened 2025-06-12
Fix up Ollama guide based on Aaishika's experience

joelhans:
@aaishikasb was building a demo around Ollama and ran into a few issues in this guide.

1. Ollama requires you to rewrite the host header to `localhost`
2. The base64 encoding wasn't correct
3. The example response from Ollama was incorrect 

This PR fixes those issues, and then I added in some general cleanup to make the steps clearer.


aaishikasb (approved):
LGTM ✅
Do you recommend adding a curl to get the message "Ollama is running"?

PR #1404, opened 2025-06-13
Update http-oauth-authn.mdx for Go SDK V2 Calling Patterns

iambmelt:
This example replaces the v1 code snippet demonstrating oauth with v2 calling patterns.

If we'd prefer to have this be a net-new snippet, not one replacing an existing code block, I'm happy to take that feedback.

### DIFF
--- a/examples/go-sdk/http-oauth-authn.mdx
+++ b/examples/go-sdk/http-oauth-authn.mdx
@@ -8,15 +8,8 @@ import (
 	"golang.ngrok.com/ngrok/v2"
 )
 
-func main() {
-	if err := run(context.Background()); err != nil {
-		log.Fatal(err)
-	}
-}
-
 const trafficPolicy = `
 on_http_request:
-  - actions:
   - name: OAuth
     actions:
       - type: oauth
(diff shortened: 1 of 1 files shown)


examples/go-sdk/http-oauth-authn.mdx
```diff
@@ -1,28 +1,55 @@
...
+	if err != nil {
+		return err
+	}
+	// Serve HTTP traffic on the ngrok endpoint
+	log.Println("Endpoint online", ln.URL())
+	return http.Serve(ln, http.HandlerFunc(handler))
 }
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprintf(w, "Hello, %s\n", r.Header.Get("authenticated-user"))
```
inconshreveable, line 49: we need an `add-header` action in the traffic policy for this to work. we used to automatically add headers in the oauth action but pulled that behavior out when we added traffic policy because it's easier to do yourself and it introduces less behavior that you'd need to turn off yourself, should look something like this

```
on_http_request:
  - actions:
  - name: OAuth
    actions:
      - type: oauth
        config:
          auth_id: oauth
          provider: google
      - type: add-headers
         config:
           headers:
             authenticated-user: "${actions.ngrok.oauth.identity.email}"
  - name: bad email
    expressions:
      - actions.ngrok.oauth.identity.email != 'alan@example.com'
    actions:
      - type: custom-response
        config:
          body: Access denied: ${actions.ngrok.oauth.identity.name}!
          status_code: 403
```

inconshreveable (approved):
lgtm minus the bit about the header.

this example is also small enough that we could remove the run/main distinction and just use log.Fatal() in `main` instead of returning an error

examples/go-sdk/http-oauth-authn.mdx
iambmelt: 849943251f59fad89f6c9740bd9371b68411c256

iambmelt:
> lgtm minus the bit about the header.
> 
> this example is also small enough that we could remove the run/main distinction and just use log.Fatal() in `main` instead of returning an error. that will be shorter and read easier as an example

db74a9b63dfe4d37776fb6535e3e9220f4948137

PR #1418, opened 2025-06-20
Create examples gallery directly from .mdx files, not YAML

joelhans:
This PR converts the examples gallery from loading via a handwritten YAML file to loading the frontmatter of the `.mdx` files directly, which means it should be easier for folks to contribute because all they have to do add the file in the `universal-gateway/examples` folder and specify one or more `categories` in the fronmatter.

But, this needs more testing and it looks like I have many type errors to fix before this goes live.


S3Prototype:
I'm trying to pare down all the PRs in this repo in preparation for the migration. I think we can reopen this PR on the new repo once it's live, and continue this work from there. It's definitely a good idea

joelhans:
@S3Prototype Agreed! I had to deprioritize this work but will definitely spent some time learning the new docs site fundamentals to bring over the same logic once it's live. I'm tired of wrangling all that YAML!

PR #1419, opened 2025-06-20
Remove outdated reference to endpoint regions

S3Prototype:
We still have documentation that refers to Region on the Endpoint eventhough that field no longer exists on the Endpoint and Endpoints are now global.


pshafton (approved):
+1

PR #1426, opened 2025-06-24
Update Agent CLI Quickstart doc

sg-writer:
part of the quickstarts project


docs/getting-started/index.mdx
```diff
@@ -42,179 +56,113 @@ curl -sSL https://ngrok-agent.s3.amazonaws.com/ngrok.asc \
...
-Forwarding                    https://84c5df474.ngrok-free.dev -> http://localhost:8080
-
-Connections                   ttl     opn     rt1     rt5     p50     p90
-                              0       0       0.00    0.00    0.00    0.00
-```
+Start up the app or service you'd like to put online.
+This is the app that your agent endpoint will forward online traffic to.
 
-Open the Forwarding URL in your browser and you will see your web application.
+If you don't have an app to work with, you can create a minimal Node.js app using the following code to set up a basic HTTP server at port `8080`.
```
sg-writer: felt like this guide needed _something_ for the user to test drive in case they don't have anything readily available so I just copied this bit from the JS SDK quickstart. We could iterate on this by providing a repo to clone but for a v1 (well, v2) I think this works as is.

docs/getting-started/index.mdx
```diff
@@ -1,20 +1,34 @@
...
 pagination_next: universal-gateway/overview
 ---
 
 import TabItem from "@theme/TabItem";
 import Tabs from "@theme/Tabs";
-import ConfigExample from "/src/components/ConfigExample.tsx";
-import SecureYourApp from "./_secure-your-app.mdx"
 
-# Quickstart
+ngrok's core functionality is built around [Agent Endpoints](/universal-gateway/agent-endpoints/): ephemeral tunnels for connecting your app, regardless of where it's running, to the ngrok platform.
```
S3Prototype:
```suggestion
ngrok's core functionality is built around [Agent Endpoints](/universal-gateway/agent-endpoints/): ephemeral tunnels that connect your locally running apps to the internet through the ngrok cloud service.
```
I think it's technically untrue to say regardless of where it's running because the agent endpoint has to run on the same machine as the app/service.

We're also abstracting a bit by saying it's connecting your app. It's forwarding traffic to a local port, but I think for the purposes of a quickstart this abstraction is fair.

I changed it to "ngrok cloud service" because I think that's a bit more specific. "ngrok platform" feels like it could mean anything, but I'm not standing by that opinion too heavily.

docs/getting-started/index.mdx
```diff
@@ -1,20 +1,34 @@
...
+One of the most common use cases is to put an app running on your local device online—for example, to share a work in progress on the public internet without hosting and publishing it.
```
S3Prototype:
```suggestion
Common use cases include granting remote access to IoT devices, enabling your Kubernetes clusters to communicate with each other, allowing external connections to a local gaming server, and more.
```

Since my suggested change for the previous sentence already mentions granting access to the internet, I thought we could cut that verbiage here and zoom in on use-cases. These are some top use cases I'm aware of. I only cut your original use case because I think these three are stronger and didn't want to have four.

docs/getting-started/index.mdx
```diff
@@ -1,20 +1,34 @@
...
 
-This quickstart will use the [ngrok agent](/agent/) to put your application on
-the internet and secure it so that only you can access it. It assumes you already have your
-own application running locally.
+The simplest way to accomplish this is to install the [ngrok Agent CLI](/agent/) and use it to create an agent endpoint that forwards traffic to your localhost.
```
S3Prototype:
```suggestion
The simplest way to accomplish this is to create an endpoint that forwards public traffic to your localhost using the [ngrok Agent CLI](/agent/).
```
I feel like putting the end goal (creating an endpoint) before mentioning the method (the CLI) makes the sentence a little less heavy, if that makes sense. I know in general that's not the case. Like, "Take the bus to the mall" is better than "Go to the mall on the bus", but when "the mall" and "the bus" are their own slightly complicated things, the latter feels easier to parse.

Thoughts?

docs/getting-started/index.mdx
```diff
@@ -1,20 +1,34 @@
...
+This quickstart walks you through that process, as well as how to implement basic security measures by requiring visitors to log in with a Google account to access your app.
 
-## 1. Install ngrok
+## What you'll need
+
+1. [An ngrok account](https://dashboard.ngrok.com/signup).
+1. Your ngrok auth token, which you can find [in the dashboard](https://dashboard.ngrok.com/get-started/your-authtoken).
```
S3Prototype, line 19: Years ago, some engineer who probably doesn't work here anymore made the environment variable NGROK_AUTHTOKEN instead of NGROK_AUTH_TOKEN and now every time I see "auth token" my brain stops for a second 🤣 

No critique here, just wanted to make you laugh

docs/getting-started/index.mdx
```diff
@@ -42,179 +56,113 @@ curl -sSL https://ngrok-agent.s3.amazonaws.com/ngrok.asc \
...
+Run this in an administrator command prompt.
 
 </TabItem>
 </Tabs>
 
-Or [download the ngrok agent from our Download
-page](https://download.ngrok.com) if you can't use one of the options above.
-
-Going to the [Setup & Installation page in the dashboard](https://dashboard.ngrok.com/get-started/setup) will also provide installation directions and commands specific to your platform.
+Or [download the ngrok agent directly](https://download.ngrok.com) if you can't use one of the options above.
```
S3Prototype:
```suggestion
Or [follow the direct installation guide](https://download.ngrok.com?tab=download) if you can't use one of the options above.
```
The link needs that query param so it'll show the relevant instructions. But also, I think the original wording needs to be worked back in because this link takes you to instructions, but I was expecting it to be a straight up download page for an executable with an install wizard or something

docs/getting-started/index.mdx
```diff
@@ -42,179 +56,113 @@ curl -sSL https://ngrok-agent.s3.amazonaws.com/ngrok.asc \
...
+## 3. Connect your account
 
-## 3. Put your app online
-
-Start ngrok by running the following command.
+Connect your agent to your ngrok account by providing your auth token as shown below—replace `$YOUR_TOKEN` with the string given to you [in the dashboard](https://dashboard.ngrok.com/get-started/your-authtoken).
 
 ```bash
-ngrok http http://localhost:8080
+ngrok config add-authtoken $YOUR_TOKEN
```
S3Prototype, line 78: When we're on Remix instead of docusaurus we'll most likely be able to do that cool thing where the user's auth token is actually in the code snippet

docs/getting-started/index.mdx
S3Prototype: If you use the LangSwitcher component (example on [this page](https://ngrok.com/docs/universal-gateway/load-balancing-multiple-clouds/#using-sdks-and-other-tools)), you could include all the example services we have in the SDK quickstarts.

Let me know if you do that and have any issues

S3Prototype (commented):
Halfway through, gonna take a break and finish. Excellent work so far

docs/getting-started/index.mdx
```diff
@@ -42,179 +56,113 @@ curl -sSL https://ngrok-agent.s3.amazonaws.com/ngrok.asc \
...
-						},
-					}
-				],
-			},
-		],
-	}}
-/>
-
-You can test your traffic policy by sending the following request to your app.
+Navigate to the directory where this file is located and run the following command to start the server:
```
S3Prototype: Minor thing, but what are our thoughts on using a colon at the end of these kinds of sentences? I've been in environments where the style was to use a period, and ones where you use a colon. Do you have a preference? I kind of prefer periods but not strongly

docs/getting-started/index.mdx
```diff
@@ -42,179 +56,113 @@ curl -sSL https://ngrok-agent.s3.amazonaws.com/ngrok.asc \
...
+          - type: oauth
+            config:
+              provider: google
+    upstream:
+      url: 8080
+      protocol: http1
+```
 
-While this guide gets you started with ngrok, you might want your tunnel to keep running even after you close your terminal window. Here are a few options to achieve this:
+Now you can start up your endpoint using its name:
```
S3Prototype, line 210:
```suggestion
:::tip
This example uses ngrok's default Google OAuth application. To use your own, see [the OAuth Traffic Policy Action documentation](https://ngrok-docs-git-sg-doc-326-agent-endpoints-cli-f7e245-ngrok-dev.vercel.app/docs/traffic-policy/actions/oauth/#google-example).
:::

Now you can start up your endpoint using its name:
```

Important to note that this example uses ngrok's google oauth config. If you want to config your own google oauth application to be used, gotta do some extra steps. Broadly speaking the ngrok google oauth app isn't useful for production use cases so people will want to know how to configure their own

S3Prototype (approved):
This is awesome. Approved to unblock, and because I trust you will handle the feedback as you see fit before merging 👍

docs/getting-started/index.mdx
sg-writer: I tend to lean on colons in this situation but I don't have a strong preference. It would be nice to standardize around one or the other as a general rule.

S3Prototype: Let's go with colons! @sg-writer

docs/getting-started/index.mdx
sg-writer: sounds good to me!

docs/getting-started/index.mdx
sg-writer: image

docs/getting-started/index.mdx
sg-writer: The Rust example doesn't really work as a single tab snippet since it involves creating a new Rust project and then editing one of the autogenerated files, and it feels weird leaving out those instructions. I've opted to just include JS, Go, and Python here, but if there's a simpler way to provide a minimal Rust script then I'm all for it.

sg-writer:
@S3Prototype I added a little conclusion in the What's Next section which I think helps tie the whole thing together. If you're into it we could consider doing the same for the other quickstarts!

PR #1434, opened 2025-06-26
Improve mobile responsiveness in examples gallery

joelhans:
Noticed this because I looked at the newsletter while on vacation, tapped through to the examples gallery, and wasn't 100% happy with what I saw. Not sure how many folks actually look at our docs on mobile but this makes the experience not perfect, but far better.


S3Prototype (approved):
Really good improvement here

PR #1438, opened 2025-06-28
Traffic Policy LLM prompt doc

kaganjd:


kaganjd:
LLM won't automatically crawl links in the doc, so still need to incorporate this dir into the txt file: https://github.com/ngrok/ngrok-docs/tree/main/traffic-policy

sidebars.js
```diff
@@ -309,6 +309,17 @@ const sidebars = {
 						},
 					],
 				},
+				{
+					label: "LLM resources",
```
S3Prototype, line 313: I think this should go above "Examples" in the sidebar hierarchy

kaganjd:
seems like we're going a different direction than sharing prompt.txt/llm.txt, going to close this for now

PR #1454, opened 2025-07-02
fix: updating all the webhook guides to use policy files

russorat:
generated with the help of Amp.

this updates all the examples in the webhook guides to use policy instead of the command line flags.


samcrichard (approved):
this is impressive

PR #1456, opened 2025-07-03
New example: Combine Authentication Methods

joelhans:
This example is based on a customer who wants to accept traffic that's authenticated via IP restrictions `OR` JWT validation. This is a much more complicated example to explain that most of what's here so far, so happy to get any feedback on how to make the human readable **What's happening here** part any clearer.

### DIFF
--- a/docs/universal-gateway/examples/combine-auth-methods.mdx
+++ b/docs/universal-gateway/examples/combine-auth-methods.mdx
@@ -46,7 +46,7 @@ You will need to change:
 
 ```yaml
 on_http_request:
-  # Check whether a request's IP matches your allow list
+  # Check whether a request's IP matches your allow list.
   - actions:
       - type: restrict-ips
         config:
@@ -58,7 +58,7 @@ on_http_request:
             - $TRUSTED_IP_FOO
             - $TRUSTED_IP_BAR
 
-  # Accept the request if its IP matches your allow list
+  # Accept the request if its IP matches your allow list.
   - expressions:
       - actions.ngrok.restrict_ips.action == "allow"
     actions:
@@ -97,7 +97,7 @@ on_http_request:
         config:
           url: https://service.internal
 
-  # Catch-all custom response for any unauthenticated requests
+  # Catch-all custom response for any unauthenticated requests.
   - actions:
       - type: custom-response
         config:
@@ -126,11 +126,11 @@ on_http_request:
 This policy has three logical sections, where the first two allow traffic to reach your upstream after authenticating with one method `OR` another, and the third is a catch-all error response for requests that fail to authenticate against any method.
 
 The first section checks every HTTP request to see if its source IP matches one on your `allow` list for `restrict-ips`, but doesn't actually enforce the restriction there.
-Instead, the following expression checks whether the [`restrict-ips` action result variable](/docs/traffic-policy/actions/restrict-ips/#action-result-variables) `actions.ngrok.restrict_ips.action` is `allow`, which means that it would have allow the request directly if you had set `enforce: true`.
-If that expression is true, then the policy forwards the request to your upstream service.
+Instead, the expression that follows the first action checks whether the [`restrict-ips` action result variable](/docs/traffic-policy/actions/restrict-ips/#action-result-variables) `actions.ngrok.restrict_ips.action` is set to `allow`, which means that it would've allowed the request directly if you had set `enforce: true`.
+If this expression is true, then the policy forwards the request to your upstream service.
 
-The second section begins the `OR` logic for all requests that did not match the IP `allow` list.
-If these requests contain a valid JWT token, then the policy forwards them to your upstream service.
+If it's false—the request didn't match the IP `allow` list—then it's instead forwarded based on whether or not it contains a valid JWT token (which may come from a couple different places).
+If it does, then the policy forwards the request to your upstream service.
 
 Finally, the policy denies all requests that did not authenticate either by IP or JWT and delivers a custom HTML response.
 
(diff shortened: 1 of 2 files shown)


docs/universal-gateway/examples/combine-auth-methods.mdx
```diff
@@ -0,0 +1,152 @@
...
+              </div>
+            </body>
+            </html>
+```
+
+**What's happening here?**
+This policy has three logical sections, where the first two allow traffic to reach your upstream after authenticating with one method `OR` another, and the third is a catch-all error response for requests that fail to authenticate against any method.
+
+The first section checks every HTTP request to see if its source IP matches one on your `allow` list for `restrict-ips`, but doesn't actually enforce the restriction there.
+Instead, the following expression checks whether the [`restrict-ips` action result variable](/docs/traffic-policy/actions/restrict-ips/#action-result-variables) `actions.ngrok.restrict_ips.action` is `allow`, which means that it would have allow the request directly if you had set `enforce: true`.
```
sg-writer: I think this helps clarify which pieces are being discussed - I had a hard time parsing the nouns and pronouns at first

```suggestion
Instead, the expression that follows the first action checks whether the [`restrict-ips` action result variable](/docs/traffic-policy/actions/restrict-ips/#action-result-variables) `actions.ngrok.restrict_ips.action` is set to `allow`, which means that it would've allowed the request directly if you had set `enforce: true`.
```

docs/universal-gateway/examples/combine-auth-methods.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+While still viewing your new Cloud Endpoint in the dashboard, copy and paste the policy below into the Traffic Policy editor.
+You will need to change:
+
+- `$TRUSTED_IP_FOO`/`$TRUSTED_IP_BAR`: Replace with public IPs of those who should have access to your service or an [IP policy](https://dashboard.ngrok.com/ip-restrictions) in your ngrok dashboard.
+- `$YOUR_JWT_DOMAIN`: The domain name for your tenant at your JWT provider—for example, with Auth0, it looks something like `https://example.us.auth0`.
+
+```yaml
+on_http_request:
+  # Check whether a request's IP matches your allow list
```
sg-writer: some comments end in a period and others don't - I'd prefer to see one or the other across the board but don't have a strong preference either way

```suggestion
  # Check whether a request's IP matches your allow list.
```

docs/universal-gateway/examples/combine-auth-methods.mdx
```diff
@@ -0,0 +1,152 @@
...
+        config:
+          # By specifying `false`, you allow the action to run and generate
+          # response variables, like `action.ngrok.restrict_ips.action, which
+          # allow you to more precisely customize how your policy responds.
+          enforce: false
+          allow:
+            - $TRUSTED_IP_FOO
+            - $TRUSTED_IP_BAR
+
+  # Accept the request if its IP matches your allow list
```
sg-writer:
```suggestion
  # Accept the request if its IP matches your allow list.
```

docs/universal-gateway/examples/combine-auth-methods.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+  # Allow the request if the JWT token is verified by your provider.
+  - expressions:
+      - actions.ngrok.jwt_validation.tokens[0].verified
+    actions:
+      - type: forward-internal
+        config:
+          url: https://service.internal
+
+  # Catch-all custom response for any unauthenticated requests
```
sg-writer:
```suggestion
  # Catch-all custom response for any unauthenticated requests.
```

docs/universal-gateway/examples/combine-auth-methods.mdx
```diff
@@ -0,0 +1,152 @@
...
+            </body>
+            </html>
+```
+
+**What's happening here?**
+This policy has three logical sections, where the first two allow traffic to reach your upstream after authenticating with one method `OR` another, and the third is a catch-all error response for requests that fail to authenticate against any method.
+
+The first section checks every HTTP request to see if its source IP matches one on your `allow` list for `restrict-ips`, but doesn't actually enforce the restriction there.
+Instead, the following expression checks whether the [`restrict-ips` action result variable](/docs/traffic-policy/actions/restrict-ips/#action-result-variables) `actions.ngrok.restrict_ips.action` is `allow`, which means that it would have allow the request directly if you had set `enforce: true`.
+If that expression is true, then the policy forwards the request to your upstream service.
```
sg-writer:
```suggestion
If this expression is true, then the policy forwards the request to your upstream service.
```

docs/universal-gateway/examples/combine-auth-methods.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+The second section begins the `OR` logic for all requests that did not match the IP `allow` list.
```
sg-writer:
```suggestion
If it's false—the request didn't match the IP `allow` list—then it's instead forwarded based on whether or not it contains a valid JWT token (which may come from a couple different places).
```

docs/universal-gateway/examples/combine-auth-methods.mdx
```diff
@@ -0,0 +1,152 @@
...
+If these requests contain a valid JWT token, then the policy forwards them to your upstream service.
```
sg-writer:
```suggestion
If it does, then the policy forwards the request to your upstream service.
```

sg-writer (commented):
Looking great overall! I took a stab at reworking the What's happening here? section a bit, hopefully my suggestions make sense!

joelhans:
@sg-writer Made all your suggested fixes... thank you! lmk if you see anything else to tweak before we merge.

PR #1459, opened 2025-07-03
fix: Extract traffic inspection into module and update guides

russorat:
takes the common text from the guides and puts them in a module so we don't have so much duplication.


S3Prototype:
There was an issue (unrelated to this PR) where the integrations hub on /docs/integrations/ was not rendering. This fixes it, but I think we should eventually move on from this structure for our integrations docs. It made more sense when we didn't have a full time docs team, but at this point it's a bit of a liability

S3Prototype (approved):

PR #1460, opened 2025-07-07
Moving mcp guide to using-ngrok-with section

S3Prototype:
This guide makes more sense in the using-ngrok-with section than in Other Guides

### DIFF
--- a/static/scripts/fix-redirect.js
+++ b/static/scripts/fix-redirect.js
@@ -1006,7 +1006,7 @@ const redirects = [
 	],
 	[
 		fromIncludes("/docs/guides/other-guides/using-mcp/"),
-		"/docs/guides/other-guides/using-mcp/",
+		"/docs/using-ngrok-with/using-mcp/",
 	],
 ];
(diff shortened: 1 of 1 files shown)


static/scripts/fix-redirect.js
```diff
@@ -1004,6 +1004,10 @@ const redirects = [
 		fromIncludes("/docs/guides/other-guides/using-tls-mutual-authentication"),
 		"/docs/agent/agent-mutual-tls-termination/",
 	],
+	[
+		fromIncludes("/docs/guides/other-guides/using-mcp/"),
+		"/docs/guides/other-guides/using-mcp/",
```
rkolavo: should this line be the new route?

S3Prototype: Good catch, that's a copy-paste error

S3Prototype:
```suggestion
		"/docs/using-ngrok-with/using-mcp/",
```

PR #1474, opened 2025-07-14
FreeBSD Does Not Have 64-bit ARM Agent

alexdlaird:
It's unclear if this is an issue with the docs, or if the engineering team intends to (but is not) publish a FreeBSD 64-bit ARM artifact of the agent. In either case, opening this as a simple docs update to raise awareness.

Note that when downloading the FreeBSD agent, no ARM-64 version is available in the dropdown: https://ngrok.com/downloads/free-bsd
Similarly, if you manually go to what would be the URL based on the patterns for all other platforms and architectures, this URL is an error 404: https://bin.equinox.io/c/bNyj1mQVY4c/ngrok-v3-stable-freebsd-arm64.tgz

As such, this PR simply removes reference to the 64-bit ARM FreeBSD agent from the docs.


S3Prototype:
@alexdlaird Sorry for the delayed response! I reached out to engineering on the day you made this PR, and it seems it's unintentional that the FreeBSD 64-bit ARM agent isn't available. The team is working on fixing this issue, but I've decided that for now your PR makes sense and we should cut mention of it from the docs until it's actually supported

S3Prototype (approved):
Thank s

alexdlaird:
> @alexdlaird Sorry for the delayed response! I reached out to engineering on the day you made this PR, and it seems it's unintentional that the FreeBSD 64-bit ARM agent isn't available. The team is working on fixing this issue, but I've decided that for now your PR makes sense and we should cut mention of it from the docs until it's actually supported

That was my guess as well, but raising this PR seemed the easiest way to raise attention to that :). Thanks!

PR #1477, opened 2025-07-17
agent: add release notes for 3.24.0

bmpngrok:


docs/agent/version-support-policy.mdx
```diff
@@ -44,7 +44,8 @@ Extended support beyond LTS is available to customers on a commercial basis. It
 
 | Support Status | Release | Release Date | Latest | Active Support | End of Life |
 | -------------- | ------- | ------------ | ------ | -------------- | ----------- |
-| Active         | 3.23x   | 2025-06-05   | 3.23.3 | 2026-06-05     | 2026-08-05  |
+| Active         | 3.24x   | 2025-07-17   | 3.24.0 | 2026-07-17     | 2026-09-17  |
+| Active         | 3.23x   | 2025-06-05   | 3.23.3 | 2026-06-20     | 2026-08-05  |
```
kmvachhani:
```suggestion
| Active         | 3.23x   | 2025-06-05   | 3.23.3 | 2026-06-26     | 2026-08-26  |
```

should this match the release date in changelog.md?

bmpngrok: Yes, _except_ that the `Release Date` column should be _the beginning of that agent line_. So your correction to the active support and EOL columns are correct (I think; depends a bit on how you read [this particular paragraph](https://www.notion.so/ngrok/Create-an-Agent-Release-5e9eeadbf2a547cc91f5b0884e306fc9?source=copy_link#13bd19ffb6548039ab0fc9feeb137120), because I can also read that as _literally only touch the `Latest` column_), but the `Release Date` column should remain 2025-June-05 for the 3.23x series

bmpngrok: (IOW you caught a mistake I made, but I can't take that fix as-is verbatim. If you tweak the release date column though I can/will)

bmpngrok: BTW, I find this table clear as mud to read for the kinds of reasons you're highlighting, but I haven't had time to think about what a better version would even look like

kmvachhani: switched release date back to 06/05, but I see what you mean on the interpretation of that paragraph could go either way, feel free to ignore the suggestion!

kmvachhani (approved):

docs/whats-new.mdx
```diff
@@ -12,6 +12,11 @@ You can expect this page to update regularly (at least monthly). We'll include t
 - Changes to our API
 - Major security fixes
 
+## July 2025
+
+- 2025-07-17 - Agent now allows configuring minimum and maximum supported TLS versions.
+- 2025-07-17 - **\[Critical Fix\]** Agent now allows enforcing mTLS certificate validation.
```
iambmelt, line 18: Maybe we link to changelog here or identify the versions so that users can glance & know without digging for affected versions. But we do have it captured somewhere... so your call

PR #1484, opened 2025-07-22
New examples: canary and blue-green deployments

joelhans:
Fairly self-explanatory! Also doing some tweaks to the sidebar ordering to promote these over Minecraft (sorry, Minecraft).

### DIFF
--- a/docs/universal-gateway/examples/blue-green-deployments.mdx
+++ b/docs/universal-gateway/examples/blue-green-deployments.mdx
@@ -17,11 +17,9 @@ You can route specific traffic to the green environment for final tests, and onc
 
 ngrok makes blue-green deployments easier by helping you:
 
-- Specify in your policy which environments are currently live and testing, and switch between them freely for each new blue-green deployment.
+- Specify in your Traffic Policy which environments are currently live and testing, and switch between them freely for each new blue-green deployment.
 - Allow requests with specific headers to access your testing environment.
-- Handle both **whole-environment** and **scoped (service-specific)** blue-green deployments:
-  - Whole-environment deployments are ideal for monolithic-ish architectures where services are coupled and need to be updated together.
-  - Scoped deployments work better when your services are entirely decoupled and you can deploy them independently.
+- Handle both <Definition meaning="Whole-environment deployments are ideal for monolithic-ish architectures where services are coupled and need to be updated together.">**whole-environment**</Definition> and <Definition meaning="Scoped deployments work better when your services are entirely decoupled and you can deploy them independently.">**scoped (service-specific)**</Definition>.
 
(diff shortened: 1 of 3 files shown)


docs/universal-gateway/examples/blue-green-deployments.mdx
```diff
@@ -0,0 +1,161 @@
...
+import Back from "./snippets/_back-to-examples.mdx";
+import TabItem from "@theme/TabItem";
+import Tabs from "@theme/Tabs";
+
+Blue-green deployments make your release processes safer and less prone to error by using an API gateway to switch production traffic between a live environment (`blue`) and a new version (`green`).
+You can route specific traffic to the green environment for final tests, and once you make the switch to green, you can roll back to blue if you find issues.
+
+ngrok makes blue-green deployments easier by helping you:
+
+- Specify in your policy which environments are currently live and testing, and switch between them freely for each new blue-green deployment.
```
S3Prototype:
```suggestion
- Specify in your Traffic Policy which environments are currently live and testing, and switch between them freely for each new blue-green deployment.
```
IMO it's worth saying the full product name

docs/universal-gateway/examples/blue-green-deployments.mdx
```diff
@@ -0,0 +1,161 @@
...
+- Allow requests with specific headers to access your testing environment.
+- Handle both **whole-environment** and **scoped (service-specific)** blue-green deployments:
+  - Whole-environment deployments are ideal for monolithic-ish architectures where services are coupled and need to be updated together.
+  - Scoped deployments work better when your services are entirely decoupled and you can deploy them independently.
```
S3Prototype:
```suggestion
- Handle both <Definition meaning="Whole-environment deployments are ideal for monolithic-ish architectures where services are coupled and need to be updated together.">**whole-environment**</Definition> and <Definition meaning="Scoped deployments work better when your services are entirely decoupled and you can deploy them independently.">**scoped (service-specific)**</Definition> blue-green deployments:
```

The definition component will allow users to hover their mouse over these terms and see these explanations, which I think is a nicer UX here. Thoughts?

S3Prototype: https://github.com/user-attachments/assets/c461f977-a3e3-4265-b425-f256261cdc58

S3Prototype (approved):
Just a few comments, but approved. I think it's time to consider putting these in sub-groups, because the list is getting pretty long. That can be a future PR though

docs/universal-gateway/examples/blue-green-deployments.mdx
joelhans: Much nicer! Thank you for that—I hadn't considered that as an option.

PR #1485, opened 2025-07-23
Update multiplexing & OR auth pattern examples

joelhans:
Per some feedback from @inconshreveable, this PR:

1. Adds wildcard domain to the multiplexing example
2. Changes the name and thus slug of the OR pattern example to be more specific
3. Adds a redirect to accommodate ☝️ 
4. Simplifies the policy to have only one expression to check which method of auth was successful (see L89)

The OR example *works*, but comes some caveats:

- The JWT part of the expression (`actions.ngrok.jwt_validation.tokens[0].verified == true`) is superfluous, as it'll always be `true` if the policy reaches this point. If JWT validation fails for any reason, the request gets an `ERR_NGROK_3408` error.
- That said, leaving that part of the expression in makes future-proof in case `jwt-validation` gets an `enforce` config.
- You can't switch the methods around, i.e. JWT validation `OR` IP restrictions, as the `jwt-validation` action can't do `enforce: false` and let you handle errors yourself.

My next takeaway is to create some product feedback to ask that every auth method supports `enforce: false` so this `OR` pattern is fully flexible.


joelhans:
~~I do need to switch this to use basic auth instead of JWT validation, so bear with me as I make one more sweep of changes!~~

Done!

S3Prototype (approved):

PR #1497, opened 2025-07-31
Suggest checking '@domain' for email checks, not just 'domain'

euank:
A check like `endsWith('example.com')` can be subverted if someone registers the domain "not-the-real-example.com", and then logs in with an email on that domain.

We should avoid suggesting something that has security issues like that.


S3Prototype (approved):
Thanks a lot for this, this is a great contribution! I noticed we had some messed up redirects so I added them onto this PR, but it's not related to your changes. Approving 👍

PR #1498, opened 2025-07-31
update owasp crs docs to include config option exclude rule ids

benjaminchan:


S3Prototype (approved):

S3Prototype:
Thank you for contributing this

PR #1528, opened 2025-08-19
Add tag frontmatter for Bot/Service User Rename

iambmelt:
Fixing this

image

### DIFF
--- a/docs/iam/bot-users.mdx
+++ b/docs/iam/bot-users.mdx
@@ -1,3 +1,9 @@
+---
+tags:
+  - Service Users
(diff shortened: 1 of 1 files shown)


S3Prototype:
Adding tags here seems not to be the solution. We can add negative weight to this result in the algolia search index, and try creating a synonym again that will make "Service User" show up when people type bot user.

We'll tackle that after the new docs site launches. Linear issue
- https://linear.app/ngrok/issue/DOC-408/make-it-so-when-you-search-bot-user-you-get-service-user

PR #1537, opened 2025-08-22
Update secrets docs -> open beta + removing 'early access' callouts

iambmelt:
I'm finding 'Developer Preview' hard to talk about in the blog post without saying 'Open Beta' to clearly indicate it's open to all. Therefore, updating this doc to use that same language. I am also removing the instructions to request Early Access, as we've now enabled this for all users and this flag is no longer needed, as well as the disclaimer about supporting a limited swath of fields.

### DIFF
--- a/docs/traffic-policy/secrets.mdx
+++ b/docs/traffic-policy/secrets.mdx
@@ -11,14 +11,11 @@ When you update a secret in a vault, it automatically rotates across all traffic
 Secrets are supported in all Traffic Policy actions and fields that support CEL.
 
 :::info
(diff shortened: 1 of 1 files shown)


docs/traffic-policy/macros/index.mdx
```diff
@@ -778,6 +778,32 @@ The following is an example of using `rand.int` with custom values:
 	}}
 />
 
+## Secrets
```
iambmelt, line 781: @sg-writer Sorry for the churn here. I'm readding this because I realized not documenting it here too actually is kinda weird!

PR #1538, opened 2025-08-22
Fixed incorrect MCP acronym

BPKelley:
The first line says MCP stands for "Messages Control Protocol", but it's actually "Model Context Protocol".

Just to confirm I wasn't thinking of the wrong thing, I double checked in the Anthropic page linked in this first line (https://support.anthropic.com/en/articles/11175166-getting-started-with-custom-connectors-using-remote-mcp), and it agrees.


S3Prototype (approved):
Thanks again!

PR #1539, opened 2025-08-22
"Bot user" to "Service user"

BPKelley:
Happened to notice an instance of the old "Bot user" term. Updated it to "Service user". 

(side note: I *highly* approve of the new name, it's much more clear)


S3Prototype:
Thank you!

PR #1542, opened 2025-08-22
Update Traffic Policy Action fields to note CEL support

sg-writer:
also refactors the terminate-tls config to use the Config components instead of a table


kate-osborn (approved):
perfect, thank you!

PR #1549, opened 2025-08-25
Fix broken redirects in `security-dev-productivity` directory

sg-writer:
IIUC this `fromIncludes` was causing all pages in this directory to redirect to the index page

https://ngrok.slack.com/archives/C09BHFQRF3M/p1756148357192719?thread_ts=1755889889.111169&cid=C09BHFQRF3M

### DIFF
--- a/static/scripts/fix-redirect.js
+++ b/static/scripts/fix-redirect.js
@@ -146,12 +146,6 @@ const redirects = [
 		"/docs/guides/other-guides/upgrade-v2-v3",
 	],
 
-	// /docs/guides/security-dev-productivity -> /docs/guides/other-guides/security-dev-productivity
-	[
-		fromIncludes("/docs/guides/security-dev-productivity"),
-		"/docs/guides/other-guides/security-dev-productivity",
-	],
-
 	// /docs/guides/securing-your-tunnels -> /docs/guides/other-guides/securing-your-tunnels
 	[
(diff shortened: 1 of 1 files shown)


static/scripts/fix-redirect.js
```diff
@@ -146,12 +146,6 @@ const redirects = [
 		"/docs/guides/other-guides/upgrade-v2-v3",
 	],
 
-	// /docs/guides/security-dev-productivity -> /docs/guides/other-guides/security-dev-productivity
-	[
-		fromIncludes("/docs/guides/security-dev-productivity"),
```
S3Prototype, line 151: @sg-writer You're right, but we could actually change this to `fromExact` to preserve the redirect and fix this issue at once (unless this redirect is unnecessary?) Approving to unblock

sg-writer: well part of the weirdness (if I'm understanding the trail correctly) is that this doc used to be located at `/docs/guides/security-dev-productivity`, then it was moved to `/docs/guides/other-guides/security-dev-productivity`, and then it was moved back to `/docs/guides/security-dev-productivity` where it currently lives. If we remove this redirect then we kill this the loop and still preserve the more recent `fromExact` with the last step in the sequence in case it's still needed. Does any of that make sense? 😅

PR #1557, opened 2025-09-03
Create how-ngrok-charges

samcrichard:
Part of packaging. Goal is to help users understand WHAT they are being charged for and HOW we define it to head off support questions.


russorat:
let me make some edits

PR #1560, opened 2025-09-04
Mark client_id field in OIDC/OAuth policies as CEL supported

kate-osborn:
We recently added CEL support to the `client_id` fields of the OIDC and OAuth actions. This PR updates the docs to reflect that change


S3Prototype (commented):
Thank you!

S3Prototype (approved):

PR #1569, opened 2025-09-09
allow ci-checks to run on forks

cody-dot-js:


cody-dot-js:
closing, we don't want to expose secrets to forks

PR #1580, opened 2025-09-12
Create odoo guide

samcrichard:
We've seen alot of inbound for this one, so it's time to make a guide! This may need some formatting help.


samcrichard (commented):
LGTM can we ship?

PR #1583, opened 2025-09-15
refactor: More 2048 Example refactoring

jonstacks:
I'm still working on migrating the 2048 examples to use the tinyllama examples. This moves all the disparate 2048 k8s manifests into 1 place so it should make the Indiana Jones style swap easier.

### DIFF
--- a/src/components/code-block.tsx
+++ b/src/components/code-block.tsx
@@ -69,6 +69,10 @@ function DocsCodeBlock({
 		metastring ? `${className} ${metastring}` : className,
 	);
 
+	if (_title) {
+		meta.title = _title;
+	}
+
 	return (
 		<CodeBlockWithInfo
 			content={children}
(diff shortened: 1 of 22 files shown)


src/components/code-block.tsx
```diff
@@ -69,6 +69,10 @@ function DocsCodeBlock({
 		metastring ? `${className} ${metastring}` : className,
 	);
 
+	if (_title) {
+		meta.title = _title;
```
jonstacks, line 73: This was broken for quite some time. Not only is this used by the triple backquote style blocks, but this ends up getting used directly when using `CodeBlock` from `@themes/CodeBlock` and it wasn't respecting the `title` prop, when it should

src/theme/CodeBlock/index.tsx
```diff
@@ -5,6 +5,12 @@ import DocsCodeBlock, { CodeBlockFallback } from "../../components/code-block";
 type Props = ComponentProps<typeof DocsCodeBlock>;
 
 export default function CodeBlock({ className, ...props }: Props) {
+	// If className is not provided, check if props.language exists. If it does,
+	// use that to construct a className, prefixed with "language-".
+	if (!className && props.language) {
```
jonstacks, line 10: Similarly, when using `CodeBlock` directly, it wasn't working when just passing the language, when it should to make it easier to use. This shims that behavior so it works correctly now.

PR #1596, opened 2025-09-21
Examples: Move examples to into sidebar sub-categories to organize them

joelhans:
I felt like the Universal Gateway -> Examples part of the sidebar was getting quite unwieldy, so I started putting them manually into different sidebar sub-folders.


sg-writer (approved):
@joelhans FYI I wouldn't invest much time into the existing docs infra at this stage—the migration away from docusaurus is now imminent and this stuff is all done differently in v2. We'll definitely be rethinking the IA across the board once the migration is done and it'd be great to bring you in to collaborate on it. That said I have no problem with merging this since you already went to the trouble of doing it! 😁

joelhans:
@sg-writer Heard! Excited to see what the new docs site is all about. Happy to pitch in on IA when the time is right!

PR #1608, opened 2025-09-30
Fix pages with fetched content

S3Prototype:
the k8s changelog and helm values pages fetch content from endpoints, but that wasn't being rendered in prod. This fixes that issue.


randseay (approved):
👍

PR #1614, opened 2025-10-01
Make definition wrapper color-mode aware

S3Prototype:
Before this change, the definition wrapper's modal was always black with white text. Now, in light mode, it's white with black text.

Test it here:
- /universal-gateway/load-balancing-multiple-clouds#using-cloud-endpoints

Before:

image

After:

image


sg-writer (approved):
Nicely done! My only suggestion (if you haven't tried it yet) would be to considering matching the line height with the main body text. But that's verging on nitpick territory!

PR #1615, opened 2025-10-01
Update rate-limit-requests.mdx

S3Prototype:
We had incorrect language in an example that said "30 requests per second" but should say per minute. The example specifies a rate of 60s, which is 60 seconds, so it should be "per minute"

### DIFF
--- a/traffic-policy/examples/rate-limit-requests.mdx
+++ b/traffic-policy/examples/rate-limit-requests.mdx
@@ -11,7 +11,7 @@ See [the `rate-limit` Traffic Policy action docs](/traffic-policy/actions/rate-l
 
 ## By endpoint
 
-This rule applies rate limiting of `30` requests per 60 seconds to the endpoint `/api/videos`.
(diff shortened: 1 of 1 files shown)


traffic-policy/examples/rate-limit-requests.mdx
```diff
@@ -11,17 +11,17 @@ See [the `rate-limit` Traffic Policy action docs](/traffic-policy/actions/rate-l
 
 ## By endpoint
 
-This rule applies rate limiting of `30` requests per second to the endpoint `/api/videos`.
+This rule applies rate limiting of `30` requests per 60 seconds to the endpoint `/api/videos`.
```
sg-writer: nitpicky but I'd drop the `backticks` from 30 and just say "per minute"

```suggestion
This rule applies rate limiting of 30 requests per minute to the endpoint `/api/videos`.
```

sg-writer (approved):

PR #1621, opened 2025-10-02
Install and configure Vale with Google Developer style guide rules

sg-writer:
WIP


package.json
```diff
@@ -9,7 +9,8 @@
     "fmt:check": "biome check . --diagnostic-level=error && prettier --check .",
     "fmt": "biome check . --write --diagnostic-level=error --organize-imports-enabled=true && prettier --write .",
     "verify-redirects": "node snippets/custom-scripts/check-redirect-conflicts.js",
-    "test-links": "mint broken-links"
+    "test-links": "mint broken-links",
+    "vale": "vale --glob='!{snippets,errors}/**' ."
```
S3Prototype, line 13: Is it possible to add vale to the dependencies? When I try to run this command I get the error that vale is not installed

Also, we should mention this command in the readme

sg-writer:
ah weird, didn't realize Vale had to be installed on your device in addition to the project itself. Thanks for addressing that.

I tried to get an LLM to address the errors and it failed miserably 🫠 so I'll try tackling that by hand.

PR #1642, opened 2025-10-20
Propose moving to using our OpenAPI schema.

samwho:
This is in response to an interaction I had on Bluesky over the weekend (https://bsky.app/profile/stephanietuerk.net/post/3m3l2veuojk2x). Our current API reference docs don't look and feel like API reference docs. I noticed Mintlify have support for OpenAPI schemas, and we have an OpenAPI schema that appears to be up to date, so this PR replaces our current reference docs with that.

I haven't gone through an updated all of the references to the old API reference docs yet, and this PR _would_ be a breaking change for anyone that's linking to our API reference docs from externally. But I really wanted to get your take on this as an idea. Should we do it? Should we not? Is there a great reason we _haven't_ done this, or why we opted to make our own API reference docs _not_ based off of the OpenAPI schema?

### DIFF
--- a/universal-gateway/tls-certificates.mdx
+++ b/universal-gateway/tls-certificates.mdx
@@ -396,7 +396,7 @@ Certificate attached to the [domain](/universal-gateway/domains) matching the
 endpoint's URL for termination.
 
 Each [domain](/universal-gateway/domains/) has an attached TLS certificate that
-is referenced by its [`certificate_id` field](/api/resources/reserved-domains).
+is referenced by its [`certificate_id` field](/api-reference/reserveddomains/get#response-certificate-id).
 If the domain object does reference any certificate, or if there is no matching
 domain object for an endpoint's URL, ngrok selects its own TLS certificate for
 termination.
@@ -483,7 +483,7 @@ bundle](#certificate-bundles) and a [private key](#private-keys). Keep in mind
 that after you create your TLS certificate, you must attach it to one or more
 domains before it will be used for termination. When working with the API, this
 is done by updating the `certificate_id` property of the [Domain API
-Resource](/api/resources/reserved-domains).
+Resource](/api-reference/reserveddomains/get#response-certificate-id).
 
 Unlike automatic TLS certificates, when you are using custom TLS certificates,
 you are responsible for managing the renewal and rotation of new certificates.


docs.json
```diff
@@ -776,131 +797,9 @@
...
-                  "api/resources/ssh-certificate-authorities",
-                  "api/resources/ssh-host-certificates",
-                  "api/resources/ssh-user-certificates"
-                ]
-              },
-              {
-                "group": "Partners",
-                "pages": ["api/resources/abuse-reports"]
+                "group": "Reference",
+                "openapi": "https://raw.githubusercontent.com/ngrok/ngrok-openapi/refs/heads/main/ngrok.yaml"
```
samwho, line 797: This is the primary change in this file. Sorry about the formatting changes, my IDE eagerly auto-formatted the file. I can revert those changes if you like.

S3Prototype:
@samwho I'm not sure what the deal is with the OpenAPI schema, so I'll look into it and see what's up there. This looks great, and we'd definitely prefer this, but maybe there's a reason we weren't using it all this time

https://ngrok.slack.com/archives/CFAT5CPBN/p1760970449724039

S3Prototype:
@samwho Based on the slack thread above, it looks great and it seems auto-updated, but it appears to lack documentation of the k8s endpoints. I've asked the eng team about it.

A compromise could be to keep the K8s docs we currently have, and just put it on a separate page from this generated stuff until those endpoints are added to the schema

S3Prototype:
Update:
image

S3Prototype:
So what remains to be done here is fixing all the broken links this created. I'll set aside some time for that in a bit, and then we can get this over the line

samwho:
@S3Prototype I went and updated the links, but now `mint broken-links` thinks the links to the OpenAPI-generated pages are broken. I've put a message about this in our shared channel with the Mintlify team.

S3Prototype:
I believe much of these conflicts are from the vale PR that cleaned up a bunch of words. Hopefully not too complex to resolve

S3Prototype:
Okay, we merged #1650, which added a custom link checker that should ignore these generated api-reference links.

Now I need to handle the merge conflicts here. Most of these are from a vale PR #1643, so they're essentially irrelevant since those pages will be deleted. But some I'm not sure why they exist.

S3Prototype (approved):
Lookin good! Thanks so much for this

PR #1643, opened 2025-10-20
Clean up all Vale errors

sg-writer:
You can test that all of the errors are cleared locally by running `pnpm run vale --minAlertLevel error`

This PR also downgrades the "Units" rule from error to warning, and we might want to consider dropping it entirely. I'm not sure it's useful.

It's possible there may be some find+replace typos in here! Hopefully I caught them all.

I also took the liberty of replacing all variations (that I could find) of `Step 1:` etc. with `1.` across the board - not technically a Vale error but I was already editing other similar headers so why not.

There are many instances where "e.g." and "i.e." were used incorrectly (which is one reason why we should avoid using them). I'm not sure if it's really worth my time to comb through them one by one to confirm "for example" and "that is" are always the best replacements—they're not exactly interchangeable but it's close.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ Several configuration files control various aspects of the documentation build p
 
 # Adding New Content
 
-- If you're adding new content (for example, a guide, example), create a new mdx file within the appropriate subdirectory of the docs directory.
+- To add new content (such as a guide or an example), create a new MDX file in the appropriate directory.
   Update the docs.json file to add the new content to the navigation sidebar if applicable.
 
 # Modifying Error Pages


S3Prototype:
Awesome stuff! I'm reviewing now. WIll take some time, but I'll try to get this over the line so we don't have a ton of conflicts bubbling up

.vale/styles/Google/Units.yml
```diff
@@ -2,7 +2,7 @@ extends: existence
 message: "Put a nonbreaking space between the number and the unit in '%s'."
 link: "https://developers.google.com/style/units-of-measure"
 nonword: true
-level: error
+level: warning
 tokens:
```
S3Prototype, line 6: I'm down with dropping this rule for now if you'd like. Feels unimportant ultimately

CONTRIBUTING.md
```diff
@@ -33,7 +33,7 @@ Several configuration files control various aspects of the documentation build p
 
 # Adding New Content
 
-- If you're adding new content (e.g., a guide, example), create a new mdx file within the appropriate subdirectory of the docs directory.
+- If you're adding new content (for example, a guide, example), create a new mdx file within the appropriate subdirectory of the docs directory.
```
S3Prototype: I agree with you on not using latin-isms (if that's what they're called?), but it'll be a rough fight. Engineers *love* e.g. 

Let's try to keep this rule in and see how others respond.

.vale/styles/Google/Units.yml
sg-writer: cool, we can take another pass at the rules to keep refining them in a separate PR

CONTRIBUTING.md
sg-writer: this one is a pretty awkward find+replace 😅 ooh and also we don't have the `docs` dir anymore

```suggestion
- To add new content (such as a guide or an example), create a new MDX file in the appropriate directory.
```

PR #1649, opened 2025-10-22
Fix TP Examples page

S3Prototype:
https://ngrok.com/docs/traffic-policy/examples

- Description is wrong
- The search component looks weird, with the search icon above the input (probably a mis-oriented flexbox)

image

This PR updates the description and removes the search icon. For some reason I can't get it to render right, so for now let's just cut it

Preview:
- https://ngrok-shaquil-doc-446-traffic-policy-example-overview-fixes.mintlify.app/traffic-policy/examples

### DIFF
--- a/snippets/ExampleHub.jsx
+++ b/snippets/ExampleHub.jsx
@@ -162,7 +162,7 @@ export const ExampleHub = ({ parentDir }) => {
 				<div className="relative max-w-64">
 					<input
 						type="text"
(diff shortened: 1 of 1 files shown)


sg-writer:
Two potential design improvements:
- make the left margin the same for "Filter..." and "All categories"
- make the "Filter..." background the same color as the main body background in dark mode

Screenshot 2025-10-22 at 11 45 11 AM
Screenshot 2025-10-22 at 11 45 00 AM

PR #1650, opened 2025-10-22
Wrap our own workflow around mintlify's broken link checker

S3Prototype:
(Surfaced [in this convo](https://ngrok.slack.com/archives/C09ETCLP0NS/p1761063300027549?thread_ts=1761038098.786619&cid=C09ETCLP0NS), related to https://github.com/ngrok/ngrok-docs/pull/1642)

Mintlify's broken-link checker doesn't work with its API reference docs generator.

Because of this, if we have links to the API reference pages that mintlify generates for us, the link checker will fail our builds because those pages do not exist in the repo when the checker runs.

Until mintlify implements a solution here, we'll have to do it ourselves.

## What this PR does

This PR adds a workflow that runs mintlify's link-checker, but doesn't fail CI if the links found are part of a list of exceptions (link-exceptions.json).

It also tests if redirects have conflicts, and fails the build if they do (prevents two different redirects with the same source)

## How to test

- You should be able to run `./custom/scripts/test-links-workflow.sh` in the root directory of the repo to test the workflow locally.
- You can try breaking a link on a random page and running the script. Should identify the broken link successfully.
- You can also try adding a link that contains one of the patterns in `link-exceptions.json` to a random file and see that it won't break, because it's an exception

## Note

This PR also moves our scripts to custom/scripts directory, out of the snippets directory, which I think makes a little more sense. No other changes have been made to those scripts that are unrelated to link and redirect checking

### DIFF
--- a/pricing-limits/free-plan-limits.mdx
+++ b/pricing-limits/free-plan-limits.mdx
@@ -38,7 +38,7 @@ You can check your usage [in the Usage page in the dashboard](https://dashboard.
 
 If you run into a limit, you can [upgrade to a plan with higher limits](https://ngrok.com/pricing?ref=docs). If you always want your endpoints online and available, select the Pay-as-you-go plan.
 
-See the [Pricing and Limits](/api-reference) page for more information.
+See the [Pricing and Limits](/pricing-limits) page for more information.
 
 ## Removing the interstitial page
 


S3Prototype:
I'm gonna push up a test commit with a broken link to make sure the CI actually fails

S3Prototype:
Okay it works. Removing the intentionally broken link now

sg-writer (approved):

PR #1653, opened 2025-10-23
Restructure api docs

S3Prototype:
This PR
- Generates the API docs into mdx files, so we can group and order them in the sidebar with `docs.json`
- Throws the "Edges" routes in a "Deprecated" group
- Adds redirects from all the old API pages to the new paths

[Preview](https://ngrok-restructure-api-docs.mintlify.app/api)

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -794,20 +794,18 @@
             "pages": [
               "api/index",
               {
-                "group": "AbuseReports",
+                "group": "API",
                 "pages": [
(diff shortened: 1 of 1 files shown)


nijikokun:
We should move backends & TunnelGroupBackends into the deprecated category as well

nijikokun:
Additionally, we should put APIKeys under API and move AbuseReports under a category called "Support"

sg-writer (approved):

PR #1671, opened 2025-10-29
Fix copy page dropdown

S3Prototype:
We have an issue where the dropdown on the "Copy page" button wasn't visible:
image

This is because we implemented a style in `style.css` that targets `button[aria-label="More actions"]` to get rid of the 3 dots in the navbar on mobile:

image

We wanted to get rid of those three dots, because if you clicked them you'd see an empty modal.

I have no idea why targeting that aria label caused the copy button dropdown to disappear, since it does not have that aria label on it so far as I can tell. But it did cause that to happen. The workaround here is to add items to the modal that appears when you click the three dots, so this PR:

1. Removes the style that hides the three dots, releasing the Copy Page dropdown from its fate
2. Adds items to the "navbar" prop in `docs.json`, which will appear in the modal for the three dots, and will sit next to the theme switcher on desktop in the navbar.

image

image

I added an item for the marketing site, and then switched our logo to go to the docs root path. I think that feels more natural anyway, and now you have a way to get back to the docs homepage that isn't awkward (clicking start took you there, but it always felt a little unintuitive that we didn't just hav esomething in the nav)

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -45,7 +45,7 @@
   "navbar": {
     "links": [
       {
-        "label": "Ngrok Homepage",
+        "label": "ngrok Homepage",
         "href": "https://ngrok.com/"
       },
       {


sg-writer (approved):
LGTM!

PR #1679, opened 2025-10-30
Add canonical URL to docs

S3Prototype:
Our ngrok.mintlify.dev URL shows up in google. It seems one thing that will help is to add ngrok.com to our canonical URL

## Relevant slack threads

- https://ngrok.slack.com/archives/C09ETCLP0NS/p1760120157307379?thread_ts=1759951003.602219&cid=C09ETCLP0NS
- https://ngrok.slack.com/archives/C09ETCLP0NS/p1760120157307379?thread_ts=1759951003.602219&cid=C09ETCLP0NS

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -2,6 +2,11 @@
   "$schema": "https://mintlify.com/docs.json",
   "theme": "mint",
   "name": "ngrok documentation",
+  "seo": {
+    "metatags": {
+        "canonical": "https://ngrok.com/docs"
+    }
+  },
   "contextual": {
     "options": [
       "copy",


docs.json
```diff
@@ -2,6 +2,11 @@
   "$schema": "https://mintlify.com/docs.json",
   "theme": "mint",
   "name": "ngrok documentation",
+  "seo": {
+    "metatags": {
+        "canonical": "https://ngrok.com/docs"
```
cody-dot-js, line 7: will this include subpaths? 

e.g.

Image

on https://ngrok.com/blog/dont-use-us-east-1

S3Prototype: According to their docs AI assistant:

> Regarding canonical URLs: When you set a canonical URL in docs.json, it applies as a base URL but does not automatically include all subpaths. The canonical tag tells search engines which domain is primary. So if you set "canonical": "https://docs.example.com", each page will have its own canonical URL like https://docs.example.com/page-path.
> 
>  If your site is hosted at /docs, yes, you should set your canonical URL to http://my-site.com/docs to ensure search engines index the correct base path for your documentation.

That answer seems... contradictory?

alex-bezek (approved):
weird i thought i had already added this https://github.com/ngrok/ngrok-docs/pull/1629/files#diff-b7b4169fa5b2e02aa17d40553a73c5d9520086753c501eeb926c63b162bb6d9eR34

docs.json
cody-dot-js: that seems fine for our use case since the "canonical origin" of docs _is_ at ngrok.com/docs NOT ngrok.com

let's try it out and adjust and/or revisit options if it's still not working

S3Prototype: Weirldy, [this PR](https://github.com/ngrok/ngrok-docs/pull/1629/files#diff-b7b4169fa5b2e02aa17d40553a73c5d9520086753c501eeb926c63b162bb6d9e) did this already. Not sure why the docs.json changes got removed, but it's worth adding back

S3Prototype:
@alex-bezek I thought so too! I guess it was overwritten somewhere between now and then 😞 But all good now

PR #1680, opened 2025-10-30
Update add headers action docs

S3Prototype:
This PR adds examples to the "behavior" section of the `add-headers` reference, to illustrate the behavior being described, where it makes sense to.

Preview: https://ngrok-shaquil-doc-455-update-add-headers-action-docs.mintlify.app/traffic-policy/actions/add-headers#behavior

### DIFF
--- a/traffic-policy/actions/add-headers.mdx
+++ b/traffic-policy/actions/add-headers.mdx
@@ -48,9 +48,9 @@ If the header already exists, it will append another header with the same key (u
 
 For example, you could use this action to add the client's IP to `x-forwarded-for`, appending information about the client making the request, so the upstream server can see where the request originated, regardless of the proxies it passed through.
 
-
 <Note>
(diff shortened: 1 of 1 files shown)


snippets/shared/phases-explanation.mdx
```diff
@@ -0,0 +1 @@
+In Traffic Policy, a phase represents a distinct point in the lifecycle of a request as it moves through an ngrok endpoint. Phases allow you to inspect, process, and manage traffic at key moments.
```
iambmelt, line 1: Why do we capitalize "Traffic Policy" but not "endpoint"? Something feels inconsistent here

iambmelt (approved):
Approving with comment

snippets/shared/phases-explanation.mdx
S3Prototype: Good callout. Vale isn't quite implemented yet, so I'll talk to SG about adding a rule here. I'm not sure if we _should_ capitalize endpoint, but I know we capitalize Traffic Policy per Niji's request

PR #1684, opened 2025-10-31
Fix pnpm fmt command

alex-bezek:
Currently if you run the `pnpm run fmt` command in the package.json file it attempts to format react code in mdx files and breaks lots of mdx views. 
Seems like we should just ignore mdx files. 

This is only a draft because i wanted to get some opinions here. If this is fine i can add this as a CI check.

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -1,0 +1,23 @@
+name: CI Checks
+
+on: push
+
+jobs:
+  ci-checks:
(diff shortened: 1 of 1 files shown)


S3Prototype:
@alex-bezek Yeah this is correct. I've not found any leads on a realistic way for it to not break codeblocks in mdx

S3Prototype:
@alex-bezek I think it'd be good to add to CI. If it's causing issues we'll just take it out later

PR #1689, opened 2025-10-31
Alex/add status info for k8s crds

jonstacks:
Updates the docs for the status conditions after the release of 0.21.0 https://github.com/ngrok/ngrok-operator/pull/697

### DIFF
--- a/k8s/crds/index.mdx
+++ b/k8s/crds/index.mdx
@@ -87,6 +87,19 @@ This document will cover the CRDs you might use to achieve your goals with the n
 
 </div>
 
+<div className="ngrok--cards ngrok--cards-flex ngrok--cards-row">
+    <Link to={`/k8s/crds/ippolicy/`}>
+        <div className="ngrok--card hover:bg-card-hover">
+            <div className="ngrok--card-heading">
+                <h3 className="text-md sm:text-base">IPPolicy</h3>
+            </div>
+            <p className="flex-grow text-md sm:text-sm">
+              A policy to restrict access to an endpoint based on IP that can be used across endpoints.
+            </p>
+        </div>
+    </Link>
+</div>
+
 ## Resource Status and Conditions
 
 The ngrok Kubernetes Operator uses standard Kubernetes conditions to report the state of resources. Each resource includes a `status.conditions` array that provides visibility into the current state and any issues that may need attention.


k8s/crds/agentendpoint.mdx
```diff
@@ -302,6 +302,316 @@ When the `spec.upstream.url` has an `https://` scheme, the client certificates w
...
+**Status Values:**
+- `True` - Agent endpoint is active and ready
+- `False` - Agent endpoint is not ready (see reason for details)
+
+**Common Reasons:**
+
+| Reason | Status | Meaning |
+|--------|--------|---------|
+| `EndpointActive` | `True` | Agent endpoint is successfully created and active |
+| `Reconciling` | `False` | Endpoint is being reconciled/updated |
```
S3Prototype: @jonstacks Does "reconciled" just mean "updated" in this context? Could we be a little more explicit about what's happening?

k8s/crds/agentendpoint.mdx
```diff
@@ -302,6 +302,316 @@ When the `spec.upstream.url` has an `https://` scheme, the client certificates w
...
+- `True` - Traffic policy is applied
+- `False` - Failed to apply traffic policy
+
+**Common Reasons:**
+
+| Reason | Status | Meaning |
+|--------|--------|---------|
+| `TrafficPolicyApplied` | `True` | Traffic policy successfully applied |
+| `TrafficPolicyError` | `False` | Error applying or validating traffic policy |
+
```
S3Prototype: This section feels like it's leaving people hanging a bit, since it says "Common Reasons" but doesn't actually list a reason there would be an error applying or validating a traffic policy. Maybe we could list one or two common issues people tend to have?

k8s/crds/agentendpoint.mdx
jonstacks: It is used when the AgentEndpoint is reconciling and not yet ready. I think we're actually going to get rid of this Reason long term, so we'll just remove it here now.

k8s/crds/agentendpoint.mdx
jonstacks: Good catch. These aren't actually common at all. I spoke with @alex-bezek and we are just going to change this to Reasons. Specifically for traffic policy, we don't have an exhaustive list of possible traffic policy errors and just pass back the error we get from ngrok.

PR #1691, opened 2025-10-31
Actually use generated event-types-table

S3Prototype:
I noticed we weren't actually using the generated _event-types-table.mdx file anywhere. I assume this wasn't intentional? It seems like it goes in the observability event logs reference, so that's where I put it.

By generating this page instead of maintaining it manually, we'll always be up to date.

This PR literally deletes all the content on the /obs/events/reference page and imports the generated content instead.

Preview:
- https://ngrok-actually-use-generated-event-types.mintlify.app/obs/events/reference

## Note

If you don't like the layout of the page, we can make edits to the ngrok go repo to change how the docs are generated. I think we should still for. now just make sure this page has accurate up to date info by importing the content.


obs/events/reference.mdx
```diff
@@ -18,7 +19,7 @@ limit which logs are captured.
 | [`http_request_complete.v0`](#http-request-completev0) | Triggers when an HTTP request completes.              |
 | [`tcp_connection_closed.v0`](#tcp-connection-closedv0) | Triggers when a TCP connection to an endpoint closes. |
 
-#### Audit Logs
+## Audit Logs
```
S3Prototype: I left these headings because they have content that was not in the _event-types-table.mdx file, but the rest of this page was duplicated in that generated page. Maybe we started generating it but forgot to stop maintaining this manually written page?

S3Prototype:
This is not quite ready yet, as the tables aren't rendered properly in some places
- Lack of headings on the tables
- Some are just broken
image

obs/events/reference.mdx
S3Prototype: Actually, this stuff was duplicated, and I've no removed these

obs/events/reference.mdx
```diff
@@ -1,1137 +1,8 @@
 ---
 title: Event Source Reference
-description: Reference documentation for all ngrok Event Sources.
+description: Reference documentation for all ngrok Log Event Sources.
```
sg-writer, line 3: nitpick: does "Log Event Sources" need to be capitalized? (I'm asking because I don't know, but it doesn't sound like it deserves proper noun treatment to me)

sg-writer (approved):
would be nice to get periods at the ends of the "Triggers when..." sentences but I'll survive without them! 😁

PR #1696, opened 2025-11-03
Generated docs with improved _event-types-tables

S3Prototype:
These generated docs have a generated event log reference that we're gonna start using in prod


errors/err_ngrok_100.mdx
```diff
@@ -1,13 +1,16 @@
...
 ---
 
 ## Message
-
 Invalid metadata length: `VAL` bytes. Max: `MAX`
 
-## Further help
 
+
+
```
sg-writer, line 13: extreme nitpick: are these extra line breaks necessary? I'm sure they don't make any difference to the rendered text but maybe worth tidying up in future iterations.

errors/err_ngrok_100.mdx
```diff
@@ -1,13 +1,16 @@
...
+## Further help
```
sg-writer, line 14: another extreme nitpick: can we add a period to the end of the sentences like this one below? 😇

PR #1697, opened 2025-11-04
Remove TLS 1.0 and 1.1 mentions in docs 

awillett31:
Remove TLS 1.0 and 1.1 mentions in docs 
https://linear.app/ngrok/project/default-minimum-tls-version-to-12-d0b9c4fb3ae2/issues

### DIFF
--- a/traffic-policy/actions/terminate-tls.mdx
+++ b/traffic-policy/actions/terminate-tls.mdx
@@ -21,11 +21,11 @@ This is the [Traffic Policy](/traffic-policy/) configuration reference for this 
 ### Configuration Fields
 
 <ConfigField title="min_version" type="string" required={false} defaultValue="1.2">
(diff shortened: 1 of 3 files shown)


traffic-policy/actions/terminate-tls.mdx
```diff
@@ -21,11 +21,11 @@ This is the [Traffic Policy](/traffic-policy/) configuration reference for this
 ### Configuration Fields
 
 <ConfigField title="min_version" type="string" required={false} defaultValue="1.2">
```
sg-writer, line 23: IIUC I think this `defaultValue` prop needs to be updated here and on line 27 as well

```suggestion
<ConfigField title="min_version" type="string" required={false} defaultValue="1.3">
```

awillett31: @sg-writer this is the min TLS version, which should be 1.2

sg-writer: My mistake!

PR #1701, opened 2025-11-04
Generated docs to fix errors

S3Prototype:
Two issues with our error docs:

## The frontmatter is rendering on the page:

[Preview](https://ngrok-ngrokrock-dev-gen-changes.mintlify.app/errors/err_ngrok_107)

Gen docs were adding extra empty lines before the frontmatter. In docusaurus, this rendered fine; in mintlify, it breaks. This PR fixes that.

### Before

image

### After

image

## Err 3200 is broken

[Preview](https://ngrok-ngrokrock-dev-gen-changes.mintlify.app/errors/err_ngrok_3200)

the `<EndpointUrl>` component used on this page is broken, and won't work in mintlify. This PR removes it. It's unnecessary and overengineered anyway. Its only purpose was to show the specific hostname of the endpoint causing the error, which would only work if a user clicked a link that had the hostname in the query params anyway. Totally unnecessary.

### Before

image

### After

image


S3Prototype:
Preview is 404-ing for some reason. 
- https://ngrok-ngrokrock-dev-gen-changes.mintlify.app/what-is-ngrok

It'll probably work if we redeploy, but I'm scared it'll break prod if we merge. Asking mintlify for help
- https://ngrok.slack.com/archives/C09ETCLP0NS/p1762267782626549

S3Prototype:
I redeployed it and it's okay. We'll see how main feels about merging this. Worst case, if it has the same error, we can redeploy main to fix it. If that doesn't work, we'll roll back this PR

sg-writer (approved):

PR #1714, opened 2025-11-06
Set up glossary generator

S3Prototype:
This PR:

- Adds a script that will generate the glossary based on the terms in the `definition-wrapper.js` script
- Adds a pre-commit hook that will run the script automatically when you make a commit that changes `definition-wrapper.js`

Preview: https://ngrok-shaquil-doc-469-set-up-the-glossary-generator.mintlify.app/glossary

## How to test

- Pull the branch
- Run `pnpm i` to install the pre-commit hook
- Edit `definition-wrapper.js`: I suggest changing the `glossaryIndex` value on line 102 to `1` or `0`. This will cause the generated glossary to be different from what we have, adding it to your files changed when the precommit hook runs
- Commit your change to `definition-wrapper.js` (make sure it will actually generate a different glossary, as described in step above)
- Check the files changed in your commit. Does it contain `glossary.mdx`? Then it's a success
- Undo your changes, so the correct glossary is generated

### DIFF
--- a/definition-wrapper.js
+++ b/definition-wrapper.js
@@ -272,7 +272,7 @@ function replaceInTextNodes(element, regex, termTitle, termObj) {
     {
       acceptNode: function(node) {
         // Skip text nodes that are inside excluded elements
-        if (isInsideExcludedElement(node.parentElement)) {
+        if (isInsideExcludedElement(node.parentElement) || window?.location?.href?.pathname("glossary")) {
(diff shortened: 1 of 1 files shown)


sg-writer (approved):
nitpick: can we turn off the definition popup on the glossary page?

Screenshot 2025-11-06 at 9 55 52 AM

S3Prototype:
@sg-writer Yes!

PR #1720, opened 2025-11-09
Replace tabs with spaces in K8s quickstart docs

joelhans:
Doing some KubeCon prep and I realized some of these YAML code blocks are broken because they use tabs instead of spaces. Trying to fix some of these ahead of KubeCon!

### DIFF
--- a/snippets/k8s/_install.mdx
+++ b/snippets/k8s/_install.mdx
@@ -37,8 +37,8 @@ The simple method only takes one command, but with the secure method, you:
 		  name: ngrok-operator-credentials
 		  namespace: ngrok-operator
 		data:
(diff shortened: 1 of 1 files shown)


jonstacks (approved):
Changes make sense to me. Still seems like its not rendering right and we probably need to figure out what the codeblock is doing to auto style it.

joelhans:
OK. @jonstacks and I figured out that the actually snippet files used in the Gateway API quickstart were moved to the root `snippets/` folder, whereas I was editing the ones in `getting-started/kubernetes/`. In a separate PR we should see if we can remove those files in `getting-started/kubernetes/` that are no longer in use.

PR #1721, opened 2025-11-10
Create GH action to run Vale as CI check

sg-writer:
This Action runs Vale as a CI check and posts comments to the PR where it finds style rule violations. If there are multiple violations on a single line it aggregates them all into one comment. It also leaves a summary comment with total error and warning count, guidance on what to do with the feedback, and a link to the style guide for more info. If there are no errors or warnings, it posts a success message instead. Each time it runs it deletes its old comments so we won't end up with duplicates when errors are not immediately addressed.

Screenshot 2025-11-14 at 12 40 42 PM

You can test it by making a change to any markdown file that breaks one of our Vale rules (like avoiding the use of I/we/us) and pushing it to this PR.

I also changed the name of the directory where our style rules live because it's been heavily customized beyond the original Google set, and it's kind of random to constantly reference Google in these comments.


sg-writer:
will it delete this even though it's not an automated comment?

S3Prototype:
I'm testing out some small tweaks. If they're not useful, I'll undo them

PR #1723, opened 2025-11-11
Restore event destinations page

S3Prototype:
Not sure if this page got lost in the migration or what, but it uses generated content, and the dashboard links to it.


S3Prototype:
Actually, now that we're gerating these docs from the OpenAPI spec, this page is no longer necessary (and should no longer be generated in the go repo)

PR #1728, opened 2025-11-12
Update nav ci check

S3Prototype:
We have a Github Action that checks if you've added or removed a file, and warns you to update the nav if you haven't. It's currently giving false positives for non-mdx files. This PR ensures only mdx files are acceptable.

## How to test

1. Pull the branch locally
2. Add a non-mdx file to one of the acceptable directories, for example `agent-sdk`
3. Push, and the "Ensure nav is updated" check should pass
4. Undo your changes, then add an empty .mdx file to an acceptable directory (the same one is fine)
5. Push, and the "ensure nav is updated" check should fail
6. Now add your new file's path to `navigation` in `docs.json` and push
7. It should pass the check
8. Do the same with deleting a random mdx file
9. Undo all your changes

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -934,7 +934,7 @@
                     "group": "Tunnels",
                     "pages": [
                       "api-reference/tunnels/list",
-                     "api-reference/tunnels/get"
+                      "api-reference/tunnels/get"
                     ]
                   },
                   {


docs.json
```diff
@@ -934,7 +934,7 @@
                     "group": "Tunnels",
                     "pages": [
                       "api-reference/tunnels/list",
-                      "api-reference/tunnels/get"
+                     "api-reference/tunnels/get"
```
sg-writer: what's up with this? intentional or just a random typo from testing?

S3Prototype: Yeah, good catch. Random typo

PR #1734, opened 2025-11-13
fix "Copy Page" dropdown

S3Prototype:


S3Prototype:
Mintlify fixed this on their end

PR #1739, opened 2025-11-14
Fix floating items in the Guides sidebar

S3Prototype:
- We have two floating items in the guides sidebar
- "Identity Aware Proxy" is a category with one item. Seems better to just slip that into security, since that doc is related to using oauth for security

image

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -544,17 +544,14 @@
                 ]
               },
               {
-                "group": "Identity Aware Proxy",
-                "pages": [
-                  "guides/identity-aware-proxy/securing-with-oauth"
-                ]
-              },
-              {
                 "group": "Security Best Practices",
                 "pages": [
                   "guides/security-dev-productivity/index",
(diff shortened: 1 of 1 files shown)


Megalonia (approved):

sg-writer (approved):
The fact that this group of pages includes multiple subdirectories makes me think that we should reevaluate how these docs are organized in general - if they all belong together then it seems like ideally we'd see a 1:1 relationship between the nav headers and the subdirectory list items (so these would all be a folder called `/security-best-practices/` for example). Beyond the scope of this PR but just wanted to point it out as part of the IA overhaul.

S3Prototype:
@sg-writer yeah, we mainly dodge moving stuff around because it makes you have to create a redirect and then find all the links. But now that we're on mintlify, and we have server-side redirects, we should be willing to move stuff to different directories so the paths change

PR #1741, opened 2025-11-17
Break out the "Guides" section in the top nav

S3Prototype:
This PR just breaks the Guides section out into its own Top Nav section and cleans up the sidebar.

It does not make any content edits. After this is merged, I'll work on adding Overviews to each new section [in a separate issue[(https://linear.app/ngrok/issue/DOC-481/un-nest-guides-section-for-better-user-experience). But I wanted to separate the content work from the restructure work here so we're not mired in two separate conversations

[Preview here](https://ngrok-shaquil-doc-481-un-nest-guides-section-for-better-use.mintlify.app/guides/api-gateway/get-started)

## How to review

I'd like this PR to largely focus on the structure of the the dropdown and the sidebar items within each Guides section. If you notice content-related issues, please comment here about them and I'll add them to a new issue to address separately

### DIFF
--- a/guides/site-to-site-connectivity/end-customers.mdx
+++ b/guides/site-to-site-connectivity/end-customers.mdx
@@ -1,6 +1,6 @@
 ---
 title: Site-to-site installation guide for end customers
-sidebarTitle: For Your Customers
+sidebarTitle: End Customer Guide
 description: Get details on why your vendor has requested you install ngrok for site-to-site connectivity, how it works, and how to maintain your end of the architecture.
 ---
 


guides/site-to-site-connectivity/end-customers.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Site-to-site installation guide for end customers
-sidebarTitle: End Customer Guide
+sidebarTitle: For Your Customers
```
nijikokun: Not sure I quite agree with this change, as it is a little more confusing.

PR #1742, opened 2025-11-17
Clean up k8s integrations and add overview page

sg-writer:
- We have some unused artifacts in k8s/integrations that need to be removed
- The k8s integration section needs an Overview page instead of defaulting to the first integration doc in the series
- Added redirects for old k8s integrations paths

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -2677,59 +2677,55 @@
       "destination": "/traffic-policy/actions/verify-webhook"
     },
     {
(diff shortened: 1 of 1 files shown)


S3Prototype:
@sg-writer Hmm. I think those pages _did_ exist prior to our migration, so I guess in this case they do need redirects. However I'll make a PR to update this CI check so it only asks you to make redirects if the file was also in docs.json. Otherwise there might be weird edge cases like this

PR #1755, opened 2025-11-20
add auto docs pr gh workflow

cody-dot-js:


.github/workflows/auto-gen-docs-pr.yml
```diff
@@ -0,0 +1,66 @@
+name: Auto-generate Docs PR
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  schedule:
+    - cron: '30 10 * * *'  # 10:30 AM UTC daily (30 min after buildkite cron)
```
S3Prototype, line 9: I actually like the idea of it being daily. We've fixed the generated docs so they're not enormous with 1000+ changes anymore, so they should be easy enough to review and get merged 👍

.github/workflows/auto-gen-docs-pr.yml
```diff
@@ -0,0 +1,84 @@
...
+        id: gh-pr
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+        run: |
+          # Try to create the PR, or update if it already exists
+          pr_url=$(gh pr create \
+            --title "Docs: Auto-generated documentation update" \
+            --body "## Auto-generated Documentation Update
+
+This PR contains automatically generated documentation changes from the ngrok monorepo.
```
S3Prototype:
```suggestion
This PR contains automatically generated documentation changes from the ngrok monorepo.

⚠️ This PR is generated daily at 10:30am [UTC](https://www.timeanddate.com/time/zone/timezone/utc)
```

I think it's worth noting how frequently this will be generated. But that makes me wonder @cody-dot-js will this run on the weekends as well?

PR #1758, opened 2025-11-20
Top nav restructure

S3Prototype:
This is a restructure that handles 90% of our planned top nav changes.

The final 10% is the "Start" and proposed "About" tabs. We need help from design to create a good landing page for the new "Start" tab so for now I'm leaving it alone.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1428,14 +1428,6 @@
         ]
       },
       {
-        "tab": "Pricing",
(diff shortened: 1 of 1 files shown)


emmamazeky-commits (commented):
b4f9ca3a5d22dc4cb0a1d079c460fe749a92d1b3

PR #1761, opened 2025-11-21
Enforce sentence case for all headers (DOC-487)

sg-writer:
This PR applies the Vale style rule enforcing sentence case across all non-title headers. I had to add quite a few exceptions to the list to handle random false positives with proper nouns. Also cleared out a few other Vale errors that made it through before the GH Action was merged.


sg-writer:
@S3Prototype I attempted to do that at the end here but it didn't work as expected. I'll try to address it in a follow-up PR since it's coming from the We rule which still needs to be applied across the board.

PR #1778, opened 2025-12-02
chore: Update node to latest LTS

jonstacks:
Update to node 24 as its the current LTS


jonstacks:
@copilot nodejs 24 does exist. Use the following information to verify:
* https://nodejs.org/en/download
* https://nodejs.org/en/blog/release/v24.0.0
* https://nodejs.org/en/about/previous-releases

jonstacks:
Nevermind, copilot isn't going to figure it out

PR #1780, opened 2025-12-02
Update spec for AGENT-138 discussion notes

iambmelt:
We decided to remove the `'` which wrapped the `filter` query param value.

[Merge this once the discussion has closed.](https://linear.app/ngrok/issue/AGENT-138/spike-on-api-filtering-formatting-for-filter-field)


api/api-filtering.mdx
```diff
@@ -17,13 +17,13 @@ To use API Filtering, you pass a subset of CEL expressions to the `filter` query
 This example request fetches a list of all your Cloud and Agent endpoints.
 
 ```http
-GET /endpoints?filter='obj.type == "cloud" || obj.type == "agent"'
+GET /endpoints?filter=obj.type == "cloud" || obj.type == "agent"
```
kmvachhani, line 20: do you think it'd be helpful if we had a full valid curl example at the beginning of the doc like this:

```
curl --location 'https://api.ngrok.com/endpoints' \
  --get \
  --data-urlencode '<filter>' \
  --header 'Ngrok-Version: 2' \
  --header 'Authorization: Bearer <token>'
```

 and then shortened these examples to just 
`filter=obj.type == "cloud" || obj.type == "agent"` ?

iambmelt: Added a `cURL` example in 569407706b6873c68f2afcdcc8e77f4f64a37a84

RE: examples. I'll leave as-is, as I think the current format illustrates the concepts nicely, at the expense of a marginal amount of extra chars

kmvachhani (approved):
lgtm, thanks for making the change!!

PR #1789, opened 2025-12-04
Document use of pagination with API Filtering

iambmelt:
Carrys foward existing pagination behaviors

### DIFF
--- a/api/api-filtering.mdx
+++ b/api/api-filtering.mdx
@@ -38,7 +38,7 @@ curl --location 'https://api.ngrok.com/endpoints' \
 
 ### Pagination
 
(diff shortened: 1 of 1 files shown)


api/api-filtering.mdx
```diff
@@ -36,6 +36,29 @@ curl --location 'https://api.ngrok.com/endpoints' \
   --header 'Authorization: Bearer <token>'
 ```
 
+### Pagination
+
+API filtering supports with pagination. You can combine the `filter` parameter with pagination parameters to page through filtered results.
```
iambmelt:
```suggestion
API filtering supports pagination. You can combine the `filter` parameter with pagination parameters to page through filtered results.
```

PR #1777, opened 2025-12-02
Demo of blue-sky IA restructure A

S3Prototype:
- Renames "Products" in the top nav to "Universal Gateway"
- Renames "Universal Gateway" in the dropdown to "What's a Universal Gateway?"
- Adds an "Endpoints" section to the dropdown
- Restructures the Universal Gateway section to better resemble how the user will be thinking about this concept when coming from the dashboard or our marketing materials

([Based on this notion doc](https://www.notion.so/ngrok/Dashboard-Study-2b5d19ffb65480998393f77ae5457d16))

Preview
- https://ngrok-test-ia-changes.mintlify.app/what-is-ngrok

Top Nav changes:
image

Universal Gateway sidebar changes:

image

Endpoints sidebar:

image

PR #1795, opened 2025-12-05
Demo of blue-sky IA restructure B

S3Prototype:
This restructure differs from https://github.com/ngrok/ngrok-docs/pull/1777 in that it breaks "Endpoints" into two separate sections:
- Agent Endpoints
- Cloud Endpoints

It still retains the "What's a Universal Gateway?" section, which serves as an overview of the universal gateway, including examples

The way it handles Agent Endpoints and Cloud Endpoints is that both contain a lot of the same content. Protocols and Bindings would be different pages under each section, but with the same content imported as snippets.

This shows the same content in multiple places, but the content for each matching page would be managed in one file in the repo (i.e., Agent Endpoints > Protocols > HTTP/S and Cloud Endpoints > Protocols > HTTP/S pull content from a single file).

## Benefits of this approach

This allows us to separate Agent and Cloud Endpoints entirely in the nav experience.

With the assumption that the overview pages would have cards that link to the other docs sections, navigation between the sections would be easy enough for users.

This version maintains the "What is a Universal Gateway" section because breaking it up completely would greatly expand the "Products" dropdown in the top nav, and there'd be a lot of pages that would have maybe one or two items in their sidebar if they were broken into dedicated sections (Domains for example).


S3Prototype:
We've decided to go with [option B here](https://github.com/ngrok/ngrok-docs/pull/1795). Closing all these branches now.

PR #1799, opened 2025-12-06
AIG

nijikokun:


ai-gateway/changelog.mdx
```diff
@@ -0,0 +1,21 @@
...
+<Update label="2025-12-15" description="EA v0.0.1">
+  ## Early Access Release
+
+  Initial Early Access release of the AI Gateway.
+
+  ### Features
+
+  - OpenAI-compatible API for routing requests to multiple AI providers
+  - Support for OpenAI, Anthropic, Google, and other providers
+  - Multi-provider failover with automatic retries
```
Alice-Lilith: nit: technically we don't actually "retry" anything per-se since we don't have a retries config yet, but we do try different models/providers/api keys

ai-gateway/concepts/api-keys.mdx
```diff
@@ -0,0 +1,131 @@
+---
+title: Provider API Keys
+description: Understanding provider API key management in the AI Gateway
+---
+
+Provider API keys authenticate requests to AI providers like OpenAI, Anthropic, and Google. The AI Gateway offers flexible key management—from simple passthrough to centralized management with automatic failover.
```
Alice-Lilith: suggestion: `The AI Gateway` -> `An ngrok AI Gateway` or `The AI Gateway action`

ai-gateway/concepts/api-keys.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+### 1. Gateway keys override SDK keys
+
+If you configure provider API keys in the gateway, those keys are used instead of any key sent by your SDK:
+
+```yaml
+providers:
+  - id: "openai"
+    api_keys:
+      - value: ${secrets.get('openai', 'gateway-key')}
```
Alice-Lilith: suggestion: I know we want to push people into this workflow, but do you think we should add a note here along the lines of "Securely load your api key from an ngrok secret instead of keeping it plain text" or something to explain the "complexity" of this macro and sell people on the usefulness of the feature?

ai-gateway/concepts/api-keys.mdx
```diff
@@ -0,0 +1,131 @@
...
+providers:
+  - id: "openai"
+    # No api_keys specified - SDK key is forwarded
+```
+
+**Why this matters**: Quick setup with no configuration changes needed.
+
+### 3. Restricted provider mode
+
+When `only_allow_configured_providers: true`, providers without configured keys are rejected:
```
Alice-Lilith: That is not really what this does. It just flat out will not allow routing to any providers  that are not configured. IE if true, and `openai` isn't in the provider list (and isn't `.disabled` if it is), then we will never route to `openai`. Keys have no impact on this.

ai-gateway/concepts/api-keys.mdx
```diff
@@ -0,0 +1,131 @@
...
+## Intelligent key selection
+
+For more control over key selection, use `api_key_selection` with CEL expressions to select keys based on runtime metrics:
+
+```yaml
+api_key_selection:
+  strategy:
+    # Prefer keys with remaining quota
+    - "ai.keys.filter(k, k.quota.remaining_requests > 100)"
+    # Fall back to keys with low error rates
```
Alice-Lilith: Should we better explain that the first strategy to return keys wins? IE: we only fall back to the second strategy if this first one doesn't return any keys?

ai-gateway/concepts/models.mdx
```diff
@@ -0,0 +1,150 @@
...
+- You want to force a specific provider
+- You're using a custom provider you've configured
+
+### With model author
+
+Use the `author/model` format when you want to be specific about who authored the model:
+
+```json
+{
+  "model": "meta-llama/llama-3-70b",
```
Alice-Lilith: nit: these specific models might not make great examples given they're not from the model catalog

ai-gateway/concepts/models.mdx
```diff
@@ -0,0 +1,150 @@
...
+{
+  "model": "gpt-4o",
+  "models": ["gpt-4o-mini", "claude-3-5-sonnet-20241022"],
+  "messages": [{"role": "user", "content": "Hello"}]
+}
+```
+
+The gateway tries models in order:
+1. `model` field first (primary choice)
+2. `models` array entries as fallbacks
```
Alice-Lilith, line 77: should we add a comment like supplying model, model+models, or even just models are all supported ways to define client model preference?

ai-gateway/concepts/models.mdx
```diff
@@ -0,0 +1,150 @@
...
+}
+```
+
+This is equivalent to omitting the model field entirely. The gateway selects from available models based on your strategy (for example, lowest latency, lowest cost).
+
+## Unknown models (pass-through)
+
+If a client requests a model that isn't in the [Model Catalog](/ai-gateway/reference/model-catalog), the gateway can still route it if:
+
+1. The model name includes a provider prefix (for example, `openai:gpt-5-preview`)
```
Alice-Lilith: This also should work with author syntax if the author is a know provider like we discussed `openai/some-new-model`

ai-gateway/concepts/models.mdx
```diff
@@ -0,0 +1,150 @@
...
+2. The provider is known and allowed by your configuration
```
Alice-Lilith: suggestion: maybe a note here about known = model catalog or custom stuff you defined in your config?

ai-gateway/concepts/models.mdx
```diff
@@ -0,0 +1,150 @@
...
+- Legacy names may map to current versions
+
+You can also define custom aliases in your provider configuration:
+
+```yaml
+providers:
+  - id: "openai"
+    models:
+      - id: "gpt-4o-2024-11-20"
+        id_aliases: ["gpt-4o", "gpt-4-latest"]
```
Alice-Lilith, line 130: note here about not being able to use an alias that is already in-use by a different model? IE `id: gemini-3-pro-preview` cannot have `gpt-4.1` as an alias because that is an id/alias already in-use by a different model.

ai-gateway/concepts/providers.mdx
```diff
@@ -0,0 +1,88 @@
...
+description: Understanding AI providers in the AI Gateway
+---
+
+Providers are the AI services that host and serve models. The AI Gateway routes your requests to providers and handles authentication, failover, and observability.
+
+## Built-in providers
+
+The AI Gateway includes built-in support for these providers:
+
+| Provider | Description |
```
Alice-Lilith, line 12: at some point we'll want to have stuff about the model catalogue generated by the main repo

ai-gateway/concepts/providers.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+## Operating modes
+
+### Passthrough mode (default)
+
+When no providers are configured, the gateway operates in **passthrough mode**:
+
+- All built-in providers are available
+- Your SDK's API key is forwarded directly to the provider
+- Automatic retries and metrics are still enabled
```
Alice-Lilith: no retries if we don't have a different model/key to try. right now do never retry a key/model combo.

ai-gateway/custom-providers/azure-openai.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+## Prerequisites
+
+- [ngrok account](https://dashboard.ngrok.com/signup) with AI Gateway access
+- [Azure OpenAI Service](https://azure.microsoft.com/en-us/products/ai-services/openai-service) resource
+- Deployed model in Azure OpenAI
+
+## Overview
+
+Azure OpenAI uses a different URL structure than OpenAI, but the AI Gateway handles the translation.
```
Alice-Lilith: not sure exactly what this means, but I don't think we do anything to the URL other than handling the base prefix like `/v1` if you leave it off the request.

ai-gateway/custom-providers/azure-openai.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+```yaml
+on_http_request:
+  - type: ai-gateway
+    config:
+      headers:
+        api-version: "2024-02-15-preview"
+      providers:
+        - id: "azure-openai"
+          base_url: "https://your-resource.openai.azure.com/openai/deployments/your-deployment"
```
Alice-Lilith: If this base path is the only real diff and all the paths after it are standard openai api compatible, disregard my comment above.

ai-gateway/custom-providers/azure-openai.mdx
```diff
@@ -0,0 +1,253 @@
...
+          api_keys:
+            - value: ${secrets.get('azure-openai', 'api-key')}
+          models:
+            - id: "gpt-4o"
+```
+
+<Note>
+Azure OpenAI requires the `api-version` header. The AI Gateway adds this to all requests.
```
Alice-Lilith: We don't do anything magical to do this, so wording might be better to make it clear via explination that it's because of the config, not some special default behaviour.

ai-gateway/custom-providers/azure-openai.mdx
```diff
@@ -0,0 +1,253 @@
...
+          base_url: "https://your-resource.openai.azure.com/openai/deployments/gpt-35-deployment"
+          api_keys:
+            - value: ${secrets.get('azure-openai', 'api-key')}
+          models:
+            - id: "gpt-3.5-turbo"
+```
+
+### Multiple regions
+
+Failover across Azure regions:
```
Alice-Lilith: These won't always automatically failover unless the clients specifically request like `models: ["azure-eastus:gpt-4o", "azure-westus:gpt-4o"]`, or if the client does just request `gpt-4o` or auto/nothing and the api key strategy selects both, then we will, but if the client specifies the provider, we won't try another unless the client says to.

TL;DR `model: "azure-eastus:gpt-4o"` will never fall back to the other region and users might find that nuance confusing with this example?

ai-gateway/custom-providers/azure-openai.mdx
```diff
@@ -0,0 +1,253 @@
...
+          models:
+            - id: "gpt-4o"
+        
+        - id: "openai"
+          api_keys:
+            - value: ${secrets.get('openai', 'api-key')}
+      
+      model_selection:
+        strategy:
+          - "ai.models.filter(m, m.provider_id == 'azure-openai')"
```
Alice-Lilith: Strategy will filter out everything other than `azure-openai:gpt-4o`. If the request fails we do not run the next strategy. We only run the next strategy if the one before it returned no allowed models (based on allowed config + what the client asked for)

ai-gateway/custom-providers/index.mdx
```diff
@@ -0,0 +1,281 @@
...
+          models:
+            - id: "llama3"
+        
+        - id: "openai"
+          api_keys:
+            - value: ${secrets.get('openai', 'api-key')}
+      
+      model_selection:
+        strategy:
+          - "ai.models.filter(m, m.provider_id == 'my-provider')"
```
Alice-Lilith, line 209: same comment as earlier regarding strategy

ai-gateway/custom-providers/index.mdx
```diff
@@ -0,0 +1,281 @@
...
+            - id: "llama3"
+        
+        - id: "inference-2"
+          base_url: "https://inference-2.internal"
+          models:
+            - id: "llama3"
+      
+      model_selection:
+        strategy:
+          - "ai.models.random()"
```
Alice-Lilith, line 240: this will be any random model, not just ones from configured providers unless you filter first by provider id or something

ai-gateway/concepts/api-keys.mdx
nijikokun: I do this in other locations, we could just port that over

ai-gateway/concepts/providers.mdx
nijikokun: Yeah, it is not hard for me to update it at the moment though

ai-gateway/overview.mdx
```diff
@@ -0,0 +1,105 @@
...
+sidebarTitle: Overview
+icon: house
+description: Route requests to AI providers with automatic failover, load balancing, and observability
+---
+
+The ngrok AI Gateway lets you route requests to AI providers like OpenAI, Anthropic, and Google through a single endpoint. If one provider fails, it automatically tries another. If one provider API key hits rate limits, it switches to another.
+
+## Why use the AI Gateway?
+
+<CardGroup cols={2}>
```
S3Prototype, line 12: Probably something to resolve in the future, but it feels weird that these don't link to anything

ai-gateway/guides/creating-endpoints.mdx
```diff
@@ -0,0 +1,227 @@
+---
+title: Creating Endpoints
+description: How to create an AI Gateway endpoint using cloud endpoints, agent endpoints, or traffic policy
+---
+
+This guide shows how to create an AI Gateway endpoint manually. For the fastest setup, use the [dashboard quickstart](/ai-gateway/quickstart) instead.
```
S3Prototype: I thought this link was gonna take me to the ngrok dashboard directly

```suggestion
<Tip>
This guide shows how to create an AI Gateway endpoint manually. For a simpler, faster setup, follow [the dashboard quickstart guide](/ai-gateway/quickstart) instead.
</Tip>
```

I think wrapping it in a tip makes it more eye-catching for people skimming, and calling it the dashboard quickstart guide makes it more obvious you're not going to the dashboard from that link

ai-gateway/guides/creating-endpoints.mdx
```diff
@@ -0,0 +1,227 @@
...
+- If one key hits rate limits, the gateway automatically tries the next one
+- If OpenAI fails entirely, requests can fail over to Anthropic
+
+<Warning>
+When you configure server-side API keys, your endpoint becomes publicly accessible. Anyone with the URL can make requests using your keys. See [Securing Your Gateway](/ai-gateway/guides/securing-endpoints) to add authorization.
+</Warning>
+
+See [Managing API Keys](/ai-gateway/guides/managing-api-keys) for detailed instructions on setting up vaults and secrets.
+
+## When to use each option
```
S3Prototype, line 215: I think this could go after "Before You Start", since it tells people which option to choose. But if we want to funnel people into the Cloud Endpoint path, it makes sense to keep it at the bottom here

ai-gateway/guides/managing-api-keys.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+```yaml
+providers:
+  - id: "openai"
+    api_keys:
+      - value: "sk-proj-abc123..."
+```
+
+<Warning>
+Inline keys are visible in traffic policy configurations. Use only for local development and testing. See [ngrok Secrets](#ngrok-secrets-recommended) for the recommended approach.
```
S3Prototype, line 26: Wait, is it not possible to use secrets locally? I think this section could use a brief explanation of _why_ someone would do this, since it introduces the possibility of accidentally using the same traffic policy later in a way that exposes the secret

ai-gateway/guides/managing-api-keys.mdx
```diff
@@ -0,0 +1,257 @@
...
+</Warning>
+
+### ngrok Vaults & Secrets (recommended)
```
S3Prototype, line 29: Would love to see this option above the inline option, since it's recommended. Also worth mentioning "You can use it for free."

ai-gateway/guides/securing-endpoints.mdx
```diff
@@ -0,0 +1,344 @@
...
+  - actions:
+      - type: ai-gateway
+        config:
+          providers:
+            - id: openai
+              api_keys:
+                - value: ${secrets.get('openai', 'api-key')}
+```
+
+## Combining with IP restrictions
```
S3Prototype, line 224: Is this related to the "IP Restrictions" feature? If not, we could rename this to something like "Restricting IPs" or "IP Whitelisting"

docs.json
```diff
@@ -223,6 +223,84 @@
...
+                  "ai-gateway/guides/securing-endpoints",
+                  "ai-gateway/guides/restricting-access",
+                  "ai-gateway/guides/model-selection-strategies",
+                  "ai-gateway/guides/modifying-requests",
+                  "ai-gateway/guides/modifying-responses",
+                  "ai-gateway/guides/error-handling"
+                ]
+              },
+              {
+                "group": "Custom Providers",
```
S3Prototype, line 268: This is a bit of a devil's advocate perspective, but I didn't immediately recognize the "Custom Providers" section as containing a list of guides for providers _other than_ our built-in providers. I initially though that this was the list of providers we supported, and found it a bit odd.

Maybe we should rename the "Concepts" section to Providers, then rename the "Providers" page to Built-in Providers, keep the "Models" and "Provider API Keys" pages there, but also put in Custom Providers as a collapsed section containing all its guides.

This is mainly from the perspective that we're granting a lot more sidebar real estate to "off-brand" options than we are to the bespoke selections, which may confuse some. Just a thought

ai-gateway/overview.mdx
S3Prototype: Maybe these are the links:

- Failover: /ai-gateway/how-it-works#failover
- One endpoint many providers - /ai-gateway/guides/configuring-providers#multi-provider-configuration
- Open AI SDK - /ai-gateway/sdks/openai
- Self hosted models - This one, I don't know

docs.json
```diff
@@ -223,6 +223,84 @@
...
             ]
           },
+          {
+            "item": "AI Gateway",
+            "pages": [
+              "ai-gateway/overview",
+              "ai-gateway/quickstart",
+              "ai-gateway/how-it-works",
+              "ai-gateway/reference/model-catalog",
+              "ai-gateway/changelog",
```
S3Prototype, line 233: We could list this here as well as in the top nav Changelogs dropdown

ai-gateway/observability/log-exporting.mdx
```diff
@@ -0,0 +1,128 @@
...
+AI Gateway-specific data is **not** currently included in logs:
+
+- Which provider handled the request
+- Model selection strategy results  
+- Retry attempts and failover decisions
+- Token counts (input/output/total)
+- Cost estimates
+- Provider API key used
+
+These AI-specific log fields are on our roadmap. Email [niji@ngrok.ai](mailto:niji@ngrok.ai) to tell us what data you need - your feedback shapes our development priorities.
```
S3Prototype, line 81: It's not super weird, but it is a _little_ weird that this is a person's email rather than some generic "info@ngrok.ai" or something

ai-gateway/reference/cel-functions.mdx
```diff
@@ -0,0 +1,547 @@
...
+```yaml
+model_selection:
+  strategy:
+    # Only models with tool calling
+    - "ai.models.filter(m, 'tool-calling' in m.supported_features)"
+    # Fall back to any model
+    - "ai.models"
+```
+
+## API key selection
```
S3Prototype, line 436: This line had `{#api-key-selection}`, but custom hashes are not supported in mintlify, so the page was broken. I pushed a commit to fix it though, all good

docs.json
nijikokun: Those providers are very important, this section is very very important. I've almost moved up multiple times, but I want people to go through the main section first.

ai-gateway/guides/managing-api-keys.mdx
nijikokun: I think it tells you below, however the line here tells you that this is more for local development.

PR #1803, opened 2025-12-08
Enforce meta descriptions for all docs (DOC-503)

sg-writer:
This PR adds a Vale rule enforcing the addition of a meta description to all docs pages for SEO purposes. I wrote the error message to give the user very basic guidance: ~150 characters with a period at the end. I also used an LLM to generate many hundreds of descriptions for pages that didn't already have them, and added periods where they were missing in existing descriptions.


S3Prototype:
@sg-writer I'm gonna have to manually review these descriptions because the LLM can get weird sometimes 🤣 but this is awesome man

PR #1825, opened 2025-12-12
Fix language names in Rust SDK getting started doc

leothorp:
References to Node.js and Python on the Rust SDK page seemed like a copy-paste error.


S3Prototype:
We were investigating why a preview deployment didn't build for this branch, but we've confirmed that this is expected behavior. Thanks for the contribution 👍

PR #1831, opened 2025-12-15
Update messages and http status codes for AI Gateway errors

kate-osborn:
Some of the AI Gateway error messages and http status codes have changed. This PR updates them to match what's in the code.

### DIFF
--- a/ai-gateway/guides/troubleshooting.mdx
+++ b/ai-gateway/guides/troubleshooting.mdx
@@ -32,7 +32,7 @@ Could not reach the AI provider. The upstream request failed.
 
 **Message**
 
-Invalid request body: {err}
+Invalid request body: `ERR`
 
 **HTTP Status**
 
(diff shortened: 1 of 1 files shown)


ai-gateway/guides/troubleshooting.mdx
```diff
@@ -32,7 +32,7 @@ The request could not be proxied to the AI provider successfully
 
 **Message**
 
-The request's body could not be parsed: {err}
+Invalid request body: {err}
```
kate-osborn: Screenshot 2025-12-15 at 1 04 35 PM

just noticed in the deployment that the "{err}" doesn't render

kate-osborn: looking at the other error reference docs, it looks like we use backticks and capitals to denote these types of values. for example, this would become:

Invalid request body: `ERR`

S3Prototype: Maybe it needs to be wrapped in backticks? But I don't know why it isn't rendering

kate-osborn: i think i'll change to `ERR` to match the other error reference docs

PR #1839, opened 2025-12-16
[FE-837] persist utmCampaign from docs --> dashboard

acrobat130:
Summary:
- when someone follows a link from an email --> docs --> dashboard, we want to persist the utmCampaign through that flow
- parse the url of the page then store the utmCammpaign param into local storage and appends it to the dashboard link when someone clicks on, hovers over, or opens the context menu on a dashboard link
- by using local storage, the utmCampaign stays consistent so if they don't immediately visit the dashboard from the provided link then it will still be tracked
- by adding a .js file to the root of this project, that script will get run on every docs page
- corresponding crusty pr: https://github.com/ngrok-private/ngrok/pull/40947

How to test:
- run docs locally `pnpm run dev`
- go to any docs page with a utm_campaign query param `http://localhost:3334/what-is-ngrok?utm_campaign=ai_gateway`
- hover over the dashboard link menu, confirm that the utm_campaign is in the url
- copy the url from the context menu, the utm_campaign should be in the url there
- clicking on the link locally you will lose the utm_campaign because the changes that preserve that on the crusty side aren't deployed yet

Ref: FE-837

### DIFF
--- a/persist-query-params.js
+++ b/persist-query-params.js
@@ -1,0 +1,102 @@
+/**
+ * Dynamic Dashboard Link UtmCampaign Parameter Forwarding with Session Persistence
+ *
+ * Automatically forwards utm_campaign query parameter from the original landing page URL
+ * to the Dashboard navigation button and persists it throughout the session.
+ *
+ */
+(function() {
+  'use strict';
+
+  // localStorage key for storing utmCampaign
+  const STORAGE_KEY = 'utmCampaign';
+  const ONE_MONTH_MS = 30 * 24 * 60 * 60 * 1000;
+  
+  /**
+   * Check if current page has a utmCampaign query param and store it in localstorage
+   */
+  function storeUtmCampaign() {
+    const params = new URLSearchParams(window.location.search);
+    const utmCampaign = params.get("utm_campaign")
+    if (utmCampaign) {
+      try {
+        localStorage.setItem(STORAGE_KEY, JSON.stringify({ utmCampaign, timestamp: Date.now() }));
+      } catch (error) {
+        // Silent error handling
+      }
+    }
+  }
+
+  /**
+   * Retrieve utmCampaign from localStorage
+   * Deletes the value if it's over 1 month old
+   * returns string
+   */
+  function getStoredUtmCampaign() {
+    try {
+      const { utmCampaign, timestamp } = JSON.parse(localStorage.getItem(STORAGE_KEY))
+
+      if (timestamp) {
+        const age = Date.now() - timestamp;
+        if (age > ONE_MONTH_MS) {
+          localStorage.removeItem(STORAGE_KEY);
+          return "";
+        }
+      }
+      return utmCampaign
+    } catch (error) {
+      return "";
+    }
+  }
+
+  /**
+   * Update the dashboard link href with the utmCampaign from localStorage as a query parameter when someone triggers a click/mouseover/contextmenu event on a dashboard link
+   */
+  function handleDashboardLinkEvent(event) {
+    const link = event.target.closest('li.navbar-link a');
+    if (!link || !link.href || !link.href.includes('dashboard.ngrok.com')) {
+      return;
+    }
+
+    storeUtmCampaign()
+
+    const storedUtmCampaign = getStoredUtmCampaign();
+    if (!storedUtmCampaign) {
+      return;
+    }
+
+    try {
+      // Parse the current dashboard URL
+      const url = new URL(link.href);
+      url.searchParams.set('utm_campaign', storedUtmCampaign);
+      link.href = url.toString();
+    } catch (error) {
+      // Silent error handling
+    }
+  }
+
+  /**
+   * Initialize: Run on DOM ready
+   */
+  function init() {
+    if (window.utmForwarderInitialized) {
+      return;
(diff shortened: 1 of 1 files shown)


persist-query-params.js
```diff
@@ -0,0 +1,128 @@
...
+
+      if (document.body) {
+        observer.observe(document.body, {
+          childList: true,
+          subtree: true
+        });
+      }
+    }
+  
+    init();
```
randseay: Is there a risk of this accidentally getting initialized multiple times? I'm not sure on the details of when an included js file is run. Does it make sense to make it idempotent to protect against that?

```ts
if (window.utmForwarderInitialized) {
  return;
}

init();
window.utmForwarderInitialized = true;

File: persist-query-params.js [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+  
+    /**
+     * Store query parameters in localStorage
+		 * params: URLSearchParams
+     */
+    function storeUtmCampaign(params) {
+			const utmCampaign = params.get("utm_campaign")
+      if (utmCampaign) {
+        try {
+          localStorage.setItem(STORAGE_KEY, utmCampaign);
```
randseay: Since `localStorage` persists so long, does it make sense to improvise some sort of expiration system? Maybe save the `value` with a `timestamp`, and then on retrieval ignore anything that is more than a certain number of days/weeks/months old?

persist-query-params.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+  
+    /**
+     * Retrieve query parameters from localStorage
+     */
+    function getStoredUtmCampaign() {
+      try {
+        return localStorage.getItem(STORAGE_KEY);
+      } catch (error) {
+        return new URLSearchParams();
```
randseay: Was it intentional to sometimes return different types? That might be confusing for consumers. Maybe it should just return `null` on error?

```
return localStorage.getItem() // returns string | null

return new URLSearchParams // returns URLSearchParams
```

persist-query-params.js
```diff
@@ -0,0 +1,128 @@
...
+
+      const navbarLinks = document.querySelectorAll('li.navbar-link a');
+
+      let dashboardLink = null;
+
+      for (const link of navbarLinks) {
+        if (link.href && link.href.includes('dashboard.ngrok.com')) {
+          dashboardLink = link;
+          break;
+        }
```
randseay: Do we want this to `break` at the first match? Would it be a problem if we have multiple links back to the dashboard?

persist-query-params.js
```diff
@@ -0,0 +1,128 @@
...
+
+      history.pushState = function() {
+        originalPushState.apply(this, arguments);
+        updateDashboardLink();
+      };
+
+      history.replaceState = function() {
+        originalReplaceState.apply(this, arguments);
+        updateDashboardLink();
+      };
```
randseay: Maybe you already explored this option, but would it simplify things to listen for a click event that rewrites the outgoing URL for `dashboard.ngrok.com` links? It might be easier than keeping the links updated, and just moves the logic to run at click time.

randseay (changes requested):
This is a resourceful bit of scripting! I have some thoughts and questions. If you've already thought them through, I apologize! Let me know if any of my thoughts are unclear.

persist-query-params.js
acrobat130: oops no this is unintentional - the original script had all of the search params saved in localstorage but we only need the utmCampaign saved - i'll update this line

persist-query-params.js
acrobat130: i'm still new to mintlify, but my understanding is that any `.js` file included in the project will get run on every page. i don't think we need to protect against that but it also wouldn't hurt. here's the mintlify docs on including custom js https://www.mintlify.com/docs/customize/custom-scripts#custom-javascript

persist-query-params.js
acrobat130: hmm yea in theory we don't need this saved after they click on the dashboard link, and i think if we can just listen for the click event like you suggested, we can delete this localstorage item completely at that point

persist-query-params.js
acrobat130: let me look into that - i think that would also make it easier to delete the localstorage item once we're done with it too

persist-query-params.js
acrobat130: we currently only have the one dashboard link but you're right - it would break if we have multiple dashboard links at some point. i'll remove the `break`

persist-query-params.js
acrobat130: ok actually i think you're right about needing a timestamp - even if we listen for click events we also need to listen for mouseover and contextmenu events (maybe others too?) so there's no clear "i'm done with with the utmCampaign" moment. so i'll go ahead and add a timestamp and invalidate it after a month

persist-query-params.js
acrobat130: changed this so that it updates the link on click/mouseover/contextmenu events

persist-query-params.js
acrobat130: in my testing it looks like init is only called on full page refresh (not when clicking around the docs) as expected, but i added your suggestion just in case

PR #1849, opened 2025-12-19
Fix code snippet example - random url assignment

iambmelt:


universal-gateway/domains.mdx
```diff
@@ -81,7 +81,7 @@ For some applications, you may not care much about the actual domain for the end
 to create and use these for endpoints. Using the value `https://${randomUrl}` in the `url` field will generate a random URL for you to use with your endpoint.
```
nijikokun:
```suggestion
to create and use these for endpoints. Using the value `https://` in the `url` field will generate a random URL for you to use with your endpoint.
```

PR #1856, opened 2025-12-29
Publish the style guide on the docs site

sg-writer:
This won't be listed anywhere in the site navigation but will be accessible for all contributors at `ngrok.com/docs/style-guide`. This should make it simpler and easier to use than its previous iteration as a Notion doc, plus we get the added benefit of version control.

Ironically this doc is going to have a ton of Vale errors and warnings since it demonstrates incorrect usage. This should be resolved by adding it to the list of stuff for Vale to ignore in package.json and vale.yml.

https://ngrok-style-guide.mintlify.app/style-guide

### DIFF
--- a/style-guide.mdx
+++ b/style-guide.mdx
@@ -2,6 +2,7 @@
 title: Style Guide
 description: Style guide and best practices for ngrok documentation.
 noSidebar: true
+noindex: true
 ---
 
 ## TL;DR


style-guide.mdx
```diff
@@ -0,0 +1,486 @@
+---
+title: Style Guide
+description: Style guide and best practices for ngrok documentation.
+noSidebar: true
+---
```
S3Prototype, line 6:
```suggestion
noindex: true
---
```
Keeps this page from showing up in search:
https://www.mintlify.com/docs/organize/hidden-pages#search,-seo,-and-ai-indexing

S3Prototype (approved):

PR #1868, opened 2026-01-08
Enrich the routing and auth Traffic Policy examples docs

joelhans:
As we near the new website, which comes with a new navigational structure, we need stronger "landing pages" for how to route and authenticate with Traffic Policy until we can build more dedicated marketing pages. Right now, we don't really have dedicated docs to "all the ways you can do auth at ngrok" and "all the ways you can route by Traffic Policy variables"—it's more separated into many individual Traffic Policy actions docs.

Very much open to thoughts from you lovely folks on whether this content can/should live somewhere else! But in the short term, it 1) adds more options and 2) makes our future website stronger.


traffic-policy/examples/route-requests.mdx
```diff
@@ -1,47 +1,1028 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+
+</CodeGroup>
+
+### A/B testing with cookies
```
sg-writer, line 441: Annoying false positive error! I gotta figure out how to fine-tune this rule.

traffic-policy/examples/add-authentication.mdx
```diff
@@ -1,61 +1,445 @@
...
+| [Basic Auth](#basic-auth) | Quick demos, internal tools |
+| [OAuth](#oauth) | User-facing apps with social login |
+| [OpenID Connect (OIDC)](#openid-connect-oidc) | Enterprise apps with corporate IdPs |
+| [JWT validation](#jwt-validation) | APIs with token-based auth |
+| [IP restrictions](#ip-restrictions) | Network-level access control |
+| [Mutual TLS (mTLS)](#mutual-tls-mtls) | Machine-to-machine, zero trust |
+
+---
+
+## Basic Auth
```
sg-writer, line 26: "Basic Auth" is a proper noun in this context, right? I'll add it to the list of exceptions.

PR #1880, opened 2026-01-13
Fix images in our integrations docs

S3Prototype:
All our integration docs images are broken in prod because they reference the wrong file paths. This PR adds the missing images and updates all the file paths.

Some of the images, as far as I can tell, the files haven't been in the repo for over a year. I assume when the guides were first written the images existed, but they're lost. I just removed those screenshots from the guides. It was only about 4 out of the hundred-plus we have, so. not a big deal.

## How to test

Your best bet is just pulling this locally and browsing through the integrations sidebar to make sure none of the pages are broken or have a broken image


S3Prototype:
@sg-writer Yeah I also don't like having all these screenshots, and we should drop them where feasible when updating these docs in the future

PR #1886, opened 2026-01-14
Fix bad robots.txt policy snippet

S3Prototype:
We had a bad policy snippet for checking for robots.txt, which @sachin20-ngrok volunteered to help fix. This PR adds his improved snippet.

### DIFF
--- a/snippets/traffic-policy/gallery/AddRobotsTxt.mdx
+++ b/snippets/traffic-policy/gallery/AddRobotsTxt.mdx
@@ -1,38 +1,57 @@
   <CodeGroup>
-    ```json policy.json
+```yaml policy.yml
+on_http_request:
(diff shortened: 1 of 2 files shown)


joelhans:
This fix is totally valid, but I do want to say that the policy, as it was, is totally valid if you attach it to an agent endpoint. It just doesn't work out a cloud endpoint because of the whole terminating thing, ya know?

sachin20-ngrok:
> This fix is totally valid, but I do want to say that the policy, as it was, is totally valid if you attach it to an agent endpoint. It just doesn't work out a cloud endpoint because of the whole terminating thing, ya know?

100% a great point. Making the distinction of when it will vs not work would be a good addition to these snippets as well.

PR #1892, opened 2026-01-15
update model catalog

Megalonia:


ai-gateway/reference/model-catalog.mdx
```diff
@@ -174,6 +174,18 @@ Inference.net provides a distributed inference network for running AI models at
...
 
+### Groq
+
+| Field | Value |
+|-------|-------|
+| **Provider ID** | `groq` |
+| **Base URL** | `https://api.groq.com/openai/v1` |
+| **Website** | [groq.com](https://groq.com/) |
+
+Groq provides ultra-fast AI inference powered by their custom LPU (Language Processing Unit) hardware, offering the fastest token generation speeds for large language models.
```
sg-writer: nit: I would avoid marketing copy in this context

```suggestion
Groq provides AI inference powered by their custom LPU (Language Processing Unit) hardware.
```

nijikokun: I would actually keep it, it's actually provably true, they are the fastest by orders of magnitudes, I think actually by 5-10x

nijikokun: Image

ai-gateway/reference/model-catalog.mdx
```diff
@@ -226,3 +238,111 @@ providers:
 
 The model catalog is updated periodically to include new models and providers. For immediate access to models not yet in the catalog, add them explicitly to your provider configuration.
 
+---
+
+## Aliases reference
+
+All model, provider, and author names are matched **case-insensitively**. For example, `gpt-4o`, `GPT-4o`, and `Gpt-4O` all resolve to the same model.
```
nijikokun: The sentence is good, i've just never read case-insensitively before and it feels wrong.

ai-gateway/reference/model-catalog.mdx
sg-writer: I don't have a strong opinion, just read more like marketing than docs content to me.

ai-gateway/reference/model-catalog.mdx
sg-writer: It is a bit clunky. IIUC you could say "...names are not case-sensitive" instead

ai-gateway/reference/model-catalog.mdx
Megalonia: I like the suggested change, but I also have no strong opinion here

PR #1929, opened 2026-02-02
update models for provider groq + remove author/model format sections

benjaminchan:
add 3 more supported groq models
update display name of gpt oss models
reorder groq models to be ordered by model id
remove author/model sections that are no longer relevant

### DIFF
--- a/ai-gateway/reference/model-catalog.mdx
+++ b/ai-gateway/reference/model-catalog.mdx
@@ -247,7 +247,7 @@ The model catalog is updated periodically to include new models and providers. F
 ---
 
 ## Aliases reference
-All model and provider names are not case-sensitive. For example, `gpt-4o`, `GPT-4o`, and `Gpt-4O` all resolve to the same model.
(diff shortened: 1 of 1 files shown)


ai-gateway/reference/model-catalog.mdx
```diff
@@ -255,7 +247,7 @@ The model catalog is updated periodically to include new models and providers. F
 ---
 
 ## Aliases reference
-All model, provider, and author names are not case-sensitive. For example, `gpt-4o`, `GPT-4o`, and `Gpt-4O` all resolve to the same model.
+All model and provider names are not case-sensitive. For example, `gpt-4o`, `GPT-4o`, and `Gpt-4O` all resolve to the same model.
```
sg-writer:
```suggestion
Model and provider names are not case-sensitive. For example, `gpt-4o`, `GPT-4o`, and `Gpt-4O` all resolve to the same model.
```

PR #1952, opened 2026-02-10
add docs for anthropic sdk, some tweaks along the way

benjaminchan:
add documentation for anthropic sdk support
aside: standardize the ai gateway subdomain for code snippets
aside: standardize openai sdk code snippets on python/ts only
manually verify that the anthropic code snippets for python/ts do work

### DIFF
--- a/ai-gateway/sdks/index.mdx
+++ b/ai-gateway/sdks/index.mdx
@@ -42,7 +42,7 @@ The pattern is the same for any SDK—just change the base URL:
 from openai import OpenAI
 
 client = OpenAI(
-    base_url="https://your-ai-subdomain.ngrok.app/v1",
+    base_url="https://your-ai-gateway.ngrok.app/v1",
     api_key="your-api-key"
 )
 ```
@@ -51,7 +51,7 @@ client = OpenAI(
 import anthropic
 
 client = anthropic.Anthropic(
-    base_url="https://your-ai-subdomain.ngrok.app",
+    base_url="https://your-ai-gateway.ngrok.app",
     api_key="your-api-key"
 )
 ```
@@ -60,7 +60,7 @@ client = anthropic.Anthropic(
 import OpenAI from "openai";
 
 const client = new OpenAI({
-  baseURL: "https://your-ai-subdomain.ngrok.app/v1",
+  baseURL: "https://your-ai-gateway.ngrok.app/v1",
   apiKey: "your-api-key",
 });
 ```
@@ -69,7 +69,7 @@ const client = new OpenAI({
 import Anthropic from "@anthropic-ai/sdk";
 
 const client = new Anthropic({
-  baseURL: "https://your-ai-subdomain.ngrok.app",
+  baseURL: "https://your-ai-gateway.ngrok.app",
   apiKey: "your-api-key",
 });
 ```
@@ -78,7 +78,7 @@ const client = new Anthropic({
 import { createOpenAI } from "@ai-sdk/openai";
 
 const openai = createOpenAI({
-  baseURL: "https://your-ai-subdomain.ngrok.app/v1",
+  baseURL: "https://your-ai-gateway.ngrok.app/v1",
   apiKey: "your-api-key",
 });
 ```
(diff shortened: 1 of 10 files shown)


ai-gateway/sdks/anthropic.mdx
```diff
@@ -0,0 +1,280 @@
...
+title: Anthropic SDK
+sidebarTitle: Anthropic SDK
+description: Use the ngrok AI Gateway with Anthropic's official SDKs.
+---
+
+<Note>
+**Prerequisite**: You need an AI Gateway endpoint before continuing. Create one using the [dashboard quickstart](/ai-gateway/quickstart) or follow the [manual setup guide](/ai-gateway/guides/creating-endpoints).
+</Note>
+
+The AI Gateway is compatible with Anthropic's official SDKs. Simply change the `baseURL` to your gateway endpoint to get automatic failover, key rotation, and observability.
```
Alice-Lilith: suggestion: `set the optional base URL configuration field in the Anthropic SDK`

ai-gateway/sdks/anthropic.mdx
```diff
@@ -0,0 +1,280 @@
...
+
+The gateway will forward requests for Anthropic SDKs unchanged in passthrough mode, making this a good option when interested accessing new Anthropic features immediately.
```
Alice-Lilith: suggestion: after this sentence maybe add something like `anthropic also provides compatability with the OpenAI API format, so you can still use your favorite claude models with AI Gateway and an openai SDK client`

ai-gateway/sdks/index.mdx
```diff
@@ -8,14 +8,17 @@ description: Connect any OpenAI-compatible SDK to the ngrok AI Gateway.
 **Prerequisite**: You need an AI Gateway endpoint before using these SDK guides. Create one using the [dashboard quickstart](/ai-gateway/quickstart) or follow the [manual setup guide](/ai-gateway/guides/creating-endpoints).
 </Note>
 
-The AI Gateway works with any SDK that supports the OpenAI API format. Just change the `baseURL` to your gateway endpoint and you're connected.
+The AI Gateway works with official and third-party SDKs. Just change the `baseURL` to your gateway endpoint and you're connected.
```
Alice-Lilith: same comment here about base url not always being `baseURL`

ai-gateway/sdks/openai.mdx
```diff
@@ -79,79 +59,6 @@ const response = await client.chat.completions.create({
 console.log(response.choices[0].message.content);
 ```
 
-```go Go highlight={12}
```
Alice-Lilith, line 82: why were some of these examples deleted?

Alice-Lilith (approved):

ai-gateway/sdks/anthropic.mdx
benjaminchan: Thinking I'll pass on this one with the thinking that this is what we're saying for the openai sdk, agree it would be more obvious.  Or could switch both cases

ai-gateway/sdks/anthropic.mdx
benjaminchan: This page is about sdk not provider, but agree I'll add a info box that points this out to clarify to user

ai-gateway/sdks/index.mdx
benjaminchan: oh this comment and above you're saying the field name isn't always baseURL verbatim?  that's a good point, will change for anthropic and openai sdks

ai-gateway/sdks/openai.mdx
benjaminchan: We had inconsistent examples.  We had some examples that had code snippets in many languages, some examples that had snippets in only one language.  So I decided to standardize all examples on just python/typescript (arbitrary choice)

ai-gateway/sdks/index.mdx
benjaminchan: updated for both openai/anthropic

ai-gateway/sdks/anthropic.mdx
benjaminchan: added and then moved existing warning down to model selection so we don't have two warnings next to each other

PR #1966, opened 2026-02-13
[agent] Use `%LocalAppData%` instead of `%HOMEPATH%\AppData\Local`

zombiezen:
`%LocalAppData%` is the environment variable that the agent uses, and could be theoretically set to a different value.


S3Prototype (approved):
Thank you

PR #1973, opened 2026-02-18
Remove definition wrapper to test performance

S3Prototype:
We're seeing layout shift all over our site, which we hypothesize is because of the definition wrapper. Let's test it by removing the script so it no longer runs, then seeing how the site performs

### DIFF
--- a/glossary.mdx
+++ b/glossary.mdx
@@ -34,7 +34,7 @@ When your create two ngrok endpoints with the same URL (and binding), those endp
 
 ## Gateway API CRD
 
-Gateway API CRDs (Custom Resource Definitions) are a set of standardized, extensible resources that manage networking configurations like routing, gateways, and traffic policies.
+Gateway API CRDs (Custom Resource Definitions) are a set of standardized, extensible resources that manage networking configurations like routing, gateways, and Traffic Policies.
 
 [Learn More](https://gateway-api.sigs.k8s.io/guides/)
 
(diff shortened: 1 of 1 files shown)


glossary.mdx
```diff
@@ -120,21 +120,21 @@ TCP KeepAlive enables TCP connections to remain active even when no data is exch
...
+## TLS Termination
 
 TLS (Transport Layer Security) termination is the process of decrypting incoming TLS traffic at a server or load balancer before passing the unencrypted traffic to internal systems.
 
 [Learn More](/universal-gateway/tls-termination/)
 
 ## Traffic Policy
 
-Traffic Policy is a configuration language that enables you to filter, match, manage, and orchestrate traffic to your endpoints. For example, you can add authentication, send custom response, rate limit traffic, and more.
+Traffic Policy is a configuration language that enables you to filter, match, manage and orchestrate traffic to your endpoints. For example, you can add authentication, send custom response, rate limit traffic, and more.
```
sg-writer:
```suggestion
Traffic Policy is a configuration language that enables you to filter, match, manage, and orchestrate traffic to your endpoints. For example, you can add authentication, send custom response, rate limit traffic, and more.
```

glossary.mdx
```diff
@@ -34,7 +34,7 @@ When your create two ngrok endpoints with the same URL (and binding), those endp
 
 ## Gateway API CRD
 
-Gateway API CRDs (Custom Resource Definitions) are a set of standardized, extensible resources that manage networking configurations like routing, gateways, and Traffic Policies.
+Gateway API CRDs (Custom Resource Definitions) are a set of standardized, extensible resources that manage networking configurations like routing, gateways, and traffic policies.
```
sg-writer:
```suggestion
Gateway API CRDs (Custom Resource Definitions) are a set of standardized, extensible resources that manage networking configurations like routing, gateways, and Traffic Policies.
```

sg-writer (approved):

PR #1980, opened 2026-02-23
Remove Prettier from the repo

sg-writer:
Prettier has been disabled since we migrated to Mintlify because it was breaking lots of stuff. I spent some time last week (#1978) trying to make it play nice with our content but the juice just ain't worth the squeeze. It would clean up 500 files and absolutely slaughter another 500. I even tried writing custom scripts to override the stuff Prettier would do wrong but it was way too janky. Not worth it! So let's get rid of it.

### DIFF
--- a/.vale/styles/StyleRules/Headings.yml
+++ b/.vale/styles/StyleRules/Headings.yml
@@ -128,6 +128,7 @@ exceptions:
   - Modern Treasury
   - MongoDB
   - MQTT
+  - Namespace
   - MQTT Explorer
   - MFA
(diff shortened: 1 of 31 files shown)


jtszalay:
We'll need to remove prettier calls from the monorepo doc gen as well.

sg-writer:
> We'll need to remove prettier calls from the monorepo doc gen as well.

@jtszalay do you know if it's actively/effectively prettying things on that side? I don't want to remove it if it's actually useful 😅

PR #1983, opened 2026-02-24
Improve navbar active underline style

sg-writer:
This PR tweaks the top-level nav bar styles to make the active underline slightly thicker than before. Subtle but effective!

BEFORE
Screenshot 2026-02-24 at 8 10 49 AM

AFTER

Screenshot 2026-02-24 at 8 22 31 AM

https://linear.app/ngrok/issue/DOC-443/improve-navbar-active-underline-style

### DIFF
--- a/style.css
+++ b/style.css
@@ -197,18 +197,6 @@ footer#footer a {
   }
 }
 
-/* Custom tab spacing for Mintlify tabs */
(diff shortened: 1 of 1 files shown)


S3Prototype:
One day we will regret using `!important` but I assume there's no way around it due to mintlify 😂

PR #1996, opened 2026-02-27
Add note to site to site guide pointing to the customer's customer guide

sg-writer:
https://linear.app/ngrok/issue/DOC-437/add-note-to-site-to-site-guides-pointing-to-the-customers-customer


joelhans:
Thanks so much for taking this @sg-writer! It's been hanging out in my Linear favorites for way, way, way too long.

PR #2061, opened 2026-03-18
Fix vale action

alex-bezek:
Fix vale CI

This curls for a release that isn't present anymore https://github.com/errata-ai 

Everything got moved to https://github.com/vale-cli/vale

There is also a vale action now, but i don't know enough about this to want to bite off refactoring this. Maybe something we can throw copilot at


alex-bezek:
prefer https://github.com/ngrok/ngrok-docs/pull/2062 over this

PR #2119, opened 2026-04-28
Update Edges page to remove implementation details

sg-writer:
Given that Edges have been deprecated since EOY '25, I think we can safely remove the overview details from this page and just leave the migration info.


S3Prototype (commented):
I think this PR makes sense, but let's get Niji's perspective here since it's an important decision

nijikokun (approved):

PR #2120, opened 2026-04-29
docs: update remaining bot user references to service users

BPKelley:
Fixes missed references in guides and snippets where text or API examples still said "bot user" / bot_users after the rename to service users.  

(found when I was cleaning up backlog items and saw PROD-28 "Update Bot User references in docs" -- it was closed, but I decided to have Claude take one last look... and here we are!)


S3Prototype (approved):
Thanks!

PR #2127, opened 2026-05-01
Remove insecure pattern documented on OAuth page

sg-writer:
https://linear.app/ngrok/issue/DOC-528/insecure-pattern-documented-on-oauth-page


S3Prototype (approved):
Good idea

PR #2139, opened 2026-05-14
Document filterable fields on API resource list pages

iambmelt:
Adds a "Filterable fields" section with the supported field list and one example expression to each list operation page

AGENT-531

### DIFF
--- a/api-reference/endpoints/list.mdx
+++ b/api-reference/endpoints/list.mdx
@@ -2,3 +2,26 @@
 openapi: get /endpoints
 description: List all agent endpoint configurations in your ngrok account with optional filtering and pagination.
 ---
+
+## Filterable fields
+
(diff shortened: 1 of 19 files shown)


api-reference/endpoints/list.mdx
```diff
@@ -1,4 +1,27 @@
 ---
 openapi: get /endpoints
 description: List all agent endpoint configurations in your ngrok account with optional filtering and pagination.
```
iambmelt:
```suggestion
description: List all endpoint configurations in your ngrok account with optional filtering and pagination.
```

This API doesn't just return agents.

iambmelt:
Ooph. This is ugly and needs a tune up

iambmelt:
Doing this the wrong way, going to instead inject this data into the service description -> then openapi spec -> then use mintlify to autorender per existing rules. Hasta la vista

PR #2141, opened 2026-05-15
docs(agent): document v3.39.2 release

bmpngrok:


Megalonia (approved):

agent/changelog.mdx
```diff
@@ -6,6 +6,10 @@ sidebarTitle: Changelog
 
 ## v3
 
+### ngrok Agent 3.39.2 \[2026-05-15\]
+
+- Various security improvements
```
S3Prototype, line 11: @bmpngrok any particular improvements worth highlighting here from an end user Pov? Just imagining from a reader's perspective they might not take much from this

bmpngrok: There were some CVEs due to the version of Go the 4/28 release was done, but that's it, so I don't think so. I honestly wrote this to sound banal because it is

PR #2144, opened 2026-05-19
chore(ngrok-operator): Add docs callouts that namespaces must exist for bound endpoints

jonstacks:
Just adding callouts that the namespaces need to exist to our guide


S3Prototype (approved):
Thanks!

PR #2145, opened 2026-05-20
RBAC page: link to pricing page

BPKelley:
The end of the RBAC page mentions pricing, but it doesn't actually link users to our pricing page. 

Not everyone knows where they should go, if they're actually interested in the "SSO & RBAC" addon. This seems like an easy opportunity to guide them towards conversion.

Less importantly: it was slightly annoying that I expected a link to pricing, and saw that sentence had 3 links... but NONE were to pricing info! (wikipedia jargon explainers, like "SSO")

### DIFF
--- a/iam/rbac.mdx
+++ b/iam/rbac.mdx
@@ -84,4 +84,4 @@ credential may listen on with ACLs.
 
 ## Pricing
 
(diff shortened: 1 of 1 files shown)


samcrichard:
might be even more helpful here to mention that its an add on outside of pay as you go, and show an image of the slider for it in checkout flow

S3Prototype (approved):

PR #2162, opened 2026-05-27
Update docs to use mintlify's AI-native features

S3Prototype:
- Adds the `<Prompt>` component, allowing users to prompt their AI tools with instructions we provide
- Expands the AI context menu options with all of mintlify's supported tools
- Also removes an unnecessary item from the secure tunnels sidebar (we had two overviews essentially)


docs.json
```diff
@@ -10,14 +10,22 @@
...
+     "copy",
+     "view",
+     "assistant",
+     "chatgpt",
+     "claude",
+     "perplexity",
+     "grok",
+     "aistudio",
+     "devin",
+     "windsurf",
```
sg-writer, line 22: I think we could safely remove these and cut down on the clutter a bit:

```suggestion
     "copy",
     "view",
     "assistant",
     "chatgpt",
     "claude",
     "perplexity",
```

docs.json
```diff
@@ -10,14 +10,22 @@
...
       "mcp",
       "cursor",
-      "vscode"
+      "vscode",
+      "devin-mcp",
```
sg-writer, line 26: this one too

```suggestion
```

sg-writer (approved):

PR #2205, opened 2026-06-30
Add ngrok video to the start page

ryan-blunden:
We're experimenting with adding the video to the marketing site home page and wanted to see if we could add it to the docs start page as well.

CleanShot 2026-07-01 at 08 23 10

Unrelated, but I'm currently getting mintlify error when running the dev server that's resolved by removing "download-pdf" from the contextual options list. I just deleted it so I could preview locally and aren't sure whether it's genuinely invalid, or just a local preview restriction for an enterprise feature.

CleanShot 2026-07-01 at 08 31 40

### DIFF
--- a/start.mdx
+++ b/start.mdx
@@ -4,7 +4,11 @@ description: ngrok makes it easy to put your apps, APIs, and services online.
 mode: center
 ---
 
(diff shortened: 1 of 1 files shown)


samcrichard (approved):

S3Prototype:
@ryan-blunden I'll investigate the issue you found!

PR #2220, opened 2026-07-06
Update AI Gateway provider models and aliases to render from source data

ryan-blunden:
This change replaces the hard-coded provider model tables in `model-catalog.mdx` with a custom component backed by model catalog data.

A new build script reads the AI Gateway provider and author catalog source files and regenerates only the embedded data section in `AIGatewayModelCatalogProvider.jsx`. The component’s rendering logic and manual doc-specific overrides are manually maintained.

The rendered catalog is intended to stay visually close to the previous docs page, but it now tracks the source catalog more closely. Provider descriptions may differ where the source data changed, though doc-specific copy overrides can still be applied in `AIGatewayModelCatalogProvider.jsx`.

To make room for fuller model ID values without wrapping, the provider model tables now comment out the `Display Name` and `Modalities` columns. This keeps the `Model ID` column readable while preserving those columns nearby in the component for easy restoration if we want them back later.

Some legacy models and aliases are still maintained manually because they do not exist in the source catalog. These are defined in `EXTRA_MODELS_BY_PROVIDER` and `EXTRA_MODEL_ALIASES_BY_PROVIDER`. Before merging, reviewers should validate the manual fallback entries in `EXTRA_MODELS_BY_PROVIDER` and `EXTRA_MODEL_ALIASES_BY_PROVIDER` to confirm they are still valid and worth including.

The regeneration workflow is intended for internal maintainers and is run by passing the build script the directory containing the AI Gateway author and provider source files.

## Why the data Is embedded in the component

When you first look at A`IGatewayModelCatalogProvider.jsx`, you’ll probably wonder why a large data blob is embedded directly inside a custom component.

Originally, the build script wrote that data to a separate component. Unfortunately, that does not work reliably in this Mint/MDX render path.

When aiGatewayModelCatalog, getProvider, or similar helpers are defined outside AIGatewayModelCatalogProvider, the page may compile successfully but then fail at runtime with errors like ReferenceError: aiGatewayModelCatalog is not defined or ReferenceError: getProvider is not defined.

Because of that, the generated catalog data is intentionally embedded inside AIGatewayModelCatalogProvider, and the small render helpers live there as well. It is not the cleanest solution, but it works.

### DIFF
--- a/snippets/ai-gateway/AIGatewayModelCatalogProvider.jsx
+++ b/snippets/ai-gateway/AIGatewayModelCatalogProvider.jsx
@@ -1,0 +1,2214 @@
+/*
+ * This file owns the rendering and helper logic for the AI Gateway model catalog UI.
+ *
+ * scripts/build-ai-gateway-model-catalog.js does not regenerate this module. It only
+ * replaces the catalog object between the generated data markers below.
+ */
+
+export const AIGatewayModelCatalogProvider = ({
+  providerId,
+  section,
+  authorId,
+}) => {
+  const aiGatewayModelCatalog =
+    //-- GENERATED PROVIDERS DATA START --//
+  {
+    "schema_version": 1,
+    "generated_from": "data",
+    "providers": [
+      {
+        "id": "openai",
+        "slug": "openai",
+        "display_name": "OpenAI",
+        "base_url": "https://api.openai.com/v1/",
+        "website": "https://openai.com/",
+        "aliases": [
+          "openAI",
+          "open-ai",
+          "open-AI"
+        ],
+        "provider_key_required": false,
+        "description": "OpenAI is an AI research organization that develops advanced artificial intelligence models, including GPT, for natural language understanding, generation, and other AI applications.",
+        "models": [
+          {
+            "id": "gpt-5.5-pro",
+            "display_name": "GPT-5.5 Pro",
+            "author_id": "openai",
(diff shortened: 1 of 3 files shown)


S3Prototype:
@ryan-blunden Nice! So would this have to be run manually going forward? If so, maybe we Could set up a precommit hook (or a GH action) that checks for changes to the model data and regenerates it automatically, so contributors don't have to remember to run this script. Would that make sense in this circumstance?

ryan-blunden:
That would be ideal, but it's something only ngrok staff can run at the moment so probably best to keep it manual for now.

snippets/ai-gateway/AIGatewayModelCatalogProvider.jsx
```diff
@@ -0,0 +1,2214 @@
...
+      {
+        alias: "moonshotai/kimi-k2-instruct-0905",
+        resolves_to: "kimi-k2",
+      },
+    ],
+  });
+
+  const EXTRA_MODEL_ALIAS_DISPLAY_NAMES = Object.freeze({
+    moonshotai: "Moonshot AI",
+  });
```
ryan-blunden, line 1784: These don't exist in the source data so have been manually added so no models or aliases are lost from the previous model-catalog.mdx version.

Ideally, they'd either be added to the source data, or just removed if they're not longer needed or relevant.

PR #2224, opened 2026-07-07
Address similar "getting started" paths for agentic traffic (and humans too)

sg-writer:
We have two very similar paths in our docs - `/start` and `/getting-started` - that provide very different user experiences and stories about what ngrok is and does. The former is our "official" Start Here page for ngrok, whereas the latter is a getting started guide for the Agent CLI specifically. According to our docs dashboard, these are the two URLs that are most commonly visited by AI assistants and crawlers, and they're both in the top 10 for all traffic in general. I suspect this leads to confusion about ngrok's purpose and usage.

This PR:
- moves the content from `/getting-started` to a new page at `gateway/agent-cli-quickstart` which better matches the materials and the existing IA in the Gateway section
- redirects `/getting-started` to `/start` so agents won't get mixed up about where to look first
- updates all internal links accordingly


S3Prototype (approved):

sg-writer:

PR #2263, opened 2026-07-28
Update BYO TLS Certificate Billing Information

AydenAdair:
Resolves CDATA-996

### Why

The BYO TLS certificates page had no billing details, and two pricing tables elsewhere in the docs advertised a $200/certificate/month rate. Our pricing has instead been updated to only charge customers ($0.27/hour) whenever they have activity on their endpoint, so they are only billed when actually using the certification. This also adds clarification around how wildcard domains are billed, confirming that additional charges will not be incurred for using a wildcard.

See:
image


S3Prototype (approved):
Thanks!

PR #2281, opened 2026-08-04
Convert numbered step headings to Steps/Step components

sg-writer:
Replaces `## 1.` / `### 2.` style step headings with Mintlify `<Steps>`/`<Step>` components across the docs. **208 files, 892 headings.**

- **Sequences that had drifted now renumber themselves** from the component: gaps (`1, 2, 4`), duplicate numbers (two `## 4.` in the same page), skips (`1, 2, 3, 4, 6`).
- **Runs crossing a `<Tabs>`/`<Tab>` boundary are split** so a step block never straddles a tab. Two shapes needed handling: tab tags indented rather than at column 0, and — in `k8s/guides/how-to/{terminate,upstream}-tls.mdx` — a run that starts outside `<Tabs>` and continues inside a `<Tab>`, where the run has to be split at the `<Tabs>` line rather than just closed.
- **Markdown stripped from titles.** `title` takes a plain string, so backticks and bold in headings would have rendered literally.

Mintlify doesn't render `Step` titles as headings, so converting a heading removes its anchor. That broke 31 links across 21 files that pointed at `#N-slug` anchors. Retargeting them to the enclosing heading isn't possible: on pages like `gateway/custom-domains.mdx` and `azure-logs-event-destination.mdx` the Steps block sits at page root with no enclosing heading, and elsewhere the nearest preceding heading is `## What you'll need`, which precedes the block rather than containing it. Either way the reader lands somewhere other than the step.

So the references name the step instead of linking to it:

- same-page links become bold step names — `(see the **Configure Zoom to send webhooks** step)`
- the 6 cross-page links keep the link and drop the dead fragment, including three `href` props on `<Card>` components in `guides/webhooks-to-on-prem/overview.mdx`

Rendering was not checked in a live Mintlify preview. The tab-split files are the ones most worth eyeballing.


S3Prototype (approved):

sg-writer:

sg-writer:
@copilot resolve the merge conflicts in this pull request. you didn't do it before.

sg-writer:

PR #2283, opened 2026-08-04
Update our webhook integration docs

S3Prototype:
Our webhook integration docs have outdated instructions, and in most places a lot of content that's unnecessary. This PR goes through the integration guides and trims down a lot of information and relies on linking to the docs for the services instead, so we don't have to maintain the accuracy. It also updates outdated information where necessary.


sg-writer (approved):
Looks good - I think we might as well delete the Hosted Hooks page since it appears to have shut down last year and you can't even get to the docs for it anymore

S3Prototype:
I think the docs just block bot access, but they are live. I did notice some places still mentioning clearbit and orbit though so i updated those

PR #2284, opened 2026-08-05
Update AI gateway model catalog

ryan-blunden:
- Add Moonshot AI and Z.ai models and provider metadata
  - Update model limits, modalities, display names, aliases, and ordering
  - Sort each provider catalog newest to oldest by release date
  - Preserve but comment out shut-down Groq models


S3Prototype (approved):
Awesome, thanks

PR #2286, opened 2026-08-05
Update JS/Python SDK pooling examples to use the builder API

melanieseltzer:
The JS and Python examples enable pooling via `ngrok.forward()`, but `forward()` doesn't support pooling in either SDK yet - the flag is silently ignored and the endpoint comes up unpooled. 

Switched both to the builder API, which does support it. Go and Rust were already correct.


S3Prototype (approved):

S3Prototype:
Thanks so much for reviewing this code 🙏

PR #2288, opened 2026-08-06
Update connect-url.mdx for connect_cas

justngrok:
Updated Let's Encrypt section to use connect_cas: host instead of the older root_cas: host

https://ngrok.com/docs/agent/config/v3#connect_cas


sg-writer (approved):
thanks @justngrok !!

PR #2293, opened 2026-08-07
Add a Platform Overview page and fold About into Platform

sg-writer:
Creates a Platform Overview in the style of the Gateway and AI Gateway overviews, and dissolves the About tab into Platform.

**Based on `gateway-absorbs-platform-primitives` (#2292), not `main`** — that PR rewrites the exact `docs.json` region this one edits, so a main-based branch would conflict. Basing here also lets the new page use post-reslug `/gateway/...` links rather than ones that only resolve via redirect. GitHub will retarget this to `main` once #2292 merges.

## What it does

- New `platform/overview.mdx`, absorbing `what-is-ngrok`, `how-ngrok-works`, and `why-ngrok`
- FAQ and Glossary move into the Platform tab, keeping `/faq` and `/glossary`
- About tab removed; the three absorbed pages redirect to `/platform/overview`

Platform is now `Overview · Identity & Access · Observability · FAQ · Glossary` — 13 pages.

Page sections: Products, Why ngrok?, What can you do with ngrok?, Gateway primitives, Account-wide, How ngrok works, What's next?

## Content sources

`what-is-ngrok` comes from **`rewrite-what-is-ngrok`**, not `main` — that branch's rewrite is the stronger version and already frames the three products cleanly. Its three stale links are corrected to current slugs here. That branch's PR can be closed in favour of this one.

`why-ngrok` had twelve H2 value props, too long to carry intact; the six strongest become a card grid. `how-ngrok-works` was a three-card router and survives as the "How ngrok works" section.

The use cases section uses the stakeholder-approved "problems we solve" taxonomy from ngrok-private/frontend#3287 — Build & test, Play, Deploy & run, Deliver & connect. Titles and descriptions are verbatim, with terminal periods added to match surrounding prose. Each entry points at the docs equivalent of its `www` target rather than the marketing page; `/use-cases/private-connectivity` → `/gateway/endpoints/internal-endpoints` is the loosest of those mappings.

The primitives are framed as Gateway's, with observability and identity split into their own account-wide section — matching where each lives in the nav now, rather than the "shared across every product" framing this restructure moved away from.

PR #2270, opened 2026-07-30
Restructure top-level IA around products and a shared Platform tab

sg-writer:
## Summary

Restructures the top-level information architecture so products and shared platform features are no longer conflated under a single "Gateway" tab.

### Navigation
- **New Platform tab** owns the shared primitives: Endpoints (agent + cloud unified), Traffic Policy, Domains & TLS, Identity & Access, Observability, ngrok Agent, and Kubernetes Operator.
- **Gateway** slimmed to the product: overview, examples, and the use-case guides (API Gateway, Device Gateway, Site-to-Site, Webhooks-to-On-Prem, Security) moved in from Guides.
- **Integrations** promoted to its own top-level tab; **Changelogs** folded into Reference.
- **Guides** tab dissolved (`/guides` → `/start`).
- Duplicate nav wirings reduced 33 → 1; merge orphans resolved.

Tab order: **Start · Platform · Gateway · AI Gateway · Share Localhost · Integrations · Reference · Pricing · About**

### Slugs & redirects (URL-stable)
- `gateway/` endpoint pages → `endpoints/`, `gateway/` domain+TLS pages → `domains/`, `guides/` use-cases → `gateway/` (85 pages).
- Redirects added for every moved slug; 574 internal links swept across 173 files; section `img/` dirs moved with their pages.

### Cleanup
- Fixed pre-existing broken links (images, relative paths, a dead anchor) and Vale errors surfaced during QA.
- Excluded `.skills/` from the docs build.

## Verification
- `mint broken-links`: **no broken links found**
- Vale: **0 errors** (3 advisory first-person CTA warnings intentionally left)
- All 757 nav pages resolve to files; no nav page references a moved slug.

## Scope / notes
- This is an IA + slug/redirect change — **no content rewrites**. Hub/overview page rewrites and use-case re-titling are planned as a follow-up (Phase 4) on a separate branch.
- Pre-existing unrelated note: `mint dev` fails a `contextual.options` schema check on `main` (config lists AI-assistant options newer than the pinned `mint`); untouched here and worth a separate fix.

PR #2299, opened 2026-08-11
Restructure the top-level IA: Platform removed, About restored, IAM & Obs embedded

sg-writer:
Supersedes #2292 and #2293 — both can be closed in favor of this. It carries all of their work plus the docs team's latest call: **there is no Platform section.**

Follow-up to #2270. Platform was framed as "shared features across products," but the truer model is that the other products *integrate Gateway's features*. So the connectivity primitives move into Gateway and their URLs move under `/gateway/` to match; the account-level pieces stop being a top-level tab and get embedded where they're actually used.

Tab order is now `Start · Gateway · AI Gateway · Share Localhost · Integrations · Reference · Pricing · About`. Gateway and Platform both lose their top-level dropdown, so `Reference` is the only tab still using one.

## What happened to Platform

- **About content** → restored `About` tab at the end of the nav, served from `/about`. It absorbs `what-is-ngrok`, `how-ngrok-works`, and `why-ngrok`, and About also holds FAQ and Glossary.
- **Identity & Access and Observability** → no longer top-level. Embedded as groups in **both** Gateway and Share Localhost, wired to the same `iam/` and `obs/` slugs. One source of truth, linked from two places, so 10 pages are deliberately double-wired.
- **Endpoints, Traffic Policy, Domains & TLS, ngrok Agent, Kubernetes Operator** → Gateway, with URLs reslugged under `/gateway/`.

## Gateway sidenav

```
gateway/overview · how-it-works · points-of-presence
▸ Endpoints (26)      ▸ Traffic Policy (60)   ▸ Domains & TLS (11)
▸ ngrok Agent (12)    ▸ Kubernetes Operator (57)
▸ API Gateway (3)     ▸ Device Gateway (16)   ▸ Site-to-Site Connectivity (19)
▸ Webhooks to On-Prem (3)  ▸ Security Best Practices (6)
▸ Examples (27)
▸ Identity & Access (6)    ▸ Observability (4)
```

## Redirects

Every namespace that moved had been migrated before, so each was already the *destination* of existing redirects. Appending inverse rules would have produced 2-hop chains and two ping-pong loops. Instead existing destinations were repointed in place, redirects that became self-referential were dropped, and every remaining chain was flattened.

| | main | this branch |
|---|---|---|
| redirects | 458 | 626 |
| chained (>1 hop) | 62 | 0 |
| max hop depth | 3 | 1 |
| loops | 0 | 0 |

## Content

The About page consolidates the three old pages. `what-is-ngrok` is taken from the **`rewrite-what-is-ngrok`** branch rather than `main` — that rewrite is the stronger version and already frames the three products cleanly; its PR can be closed too. `why-ngrok`'s twelve value props condense into a six-card grid.

Its use cases section uses the stakeholder-approved "problems we solve" taxonomy from ngrok-private/frontend#3287, with titles and descriptions verbatim and each entry pointing at the docs equivalent of its `www` target.

## Also in here

- Endpoints landing page retitled as an overview, so the sidebar reads `Endpoints › Overview`; the duplicate `cloud-endpoints/endpoints` page removed — it rendered the same shared snippet as `endpoints/index` with `noindex: true`, and nothing linked to it.
- Pre-existing fixes: a double-slash link in three files, a link to a directory with no index page, a dead redirect target that never existed, three live nav pages whose old URLs resolved to a section anchor or another product, and a relative link that escaped its tree during a move.
- ⚠️ Two CI workflow changes, splittable if preferred. `is-nav-updated` matched nav paths by bare substring, so any prefix move reported a deleted path as still present; it now matches complete JSON strings. `Run Vale` is marked `continue-on-error` because reviewdog fetches the PR diff and GitHub 406s past 300 changed files — #2282 fixed the 20k-line cap but not this one.


sg-writer:

S3Prototype (approved):
LGTM

sg-writer:

PR #2316, opened 2026-08-20
Overhaul Gateway sidebar to promote use cases and examples, mirror dashboard structure

sg-writer:
This PR restructures the Gateway sidebar nav to bring **Use Cases** and **Examples** up below the **Getting Started** section, to help users figure out what to do next after they've learned the basics and gotten set up. From there, the header structure of the nav bar mirrors the terminology and organization of the product dashboard - that way when a user moves between the two surfaces, they will know exactly where they need to go in both places to find what they need.

## In scope

- overhaul all headers and subheaders, some page titles in Gateway sidebar
- first pass on aligning entry flow with AI Gateway - currently the stronger set of docs
- create new "buckets" for organizing the **Examples** section to cut down on clutter

## Out of scope

- content changes - this is about the sidebar experience only - we'll continue to dig deeper in later iterations
- URL changes - we've made too many of these lately and Google is unhappy about it 😅 - thankfully this PR doesn't require any
- changes to Examples sections - when the Traffic Policy docs were created they were in their own top-level section (parallel to Gateway and AI Gateway). Now the TP docs are folded into Gateway, but TP and Gateway both have their own Examples sections which is maybe not intuitive. Should they be merged? If so I'd like to address this in a follow-up.

## How to review

Peep the Gateway sidebar in the deploy preview: https://ngrok-gateway-sidebar-variant-d.mintlify.site/gateway/overview

In this round of changes my primary concern is the overall shape (as opposed to the finer details) of the sidebar - do you like the order in which information is presented? Are you happy with the new "buckets" for Examples (these are certainly open for debate, and we can also continue to iterate on them)? I'm less concerned with the order of the pages _within_ individual sections - unless you see something glaring that needs to be addressed, we can do another round of polishing (if needed) in a follow-up PR.


joelhans:
Let's definitely chat about the verbiage/messaging about this as next pass... maybe next week? I'd be happy to take the lead on creating the PR and having you review it instead of having all that land on you.

Love this work overall. I like having the use cases up high and I think you've done a great job organizing all my examples into 4 categories that make a ton of sense. As an aside, we should probably also chat about how to breathe some more life into the example gallery, but that's another thing.

One big thought:

Under the **Use Cases** section, the individual categories get a little muddy. There's a lot of items here, and the titles are perhaps not the most useful, whereas before, they had their own sections that kind of explained what type of content they are. Here, I'm like, _What does **Linux** mean?_

Screenshot 2026-08-25 at 12 09 13 PM

The way it is now, you can see that Linux is in the Installation Guides section, so it all makes sense.

Screenshot 2026-08-25 at 12 14 11 PM

I truly don't know if Mintlify lets you do any kind of organization _within_ these sections, but it would be nice!

sg-writer:
> Here, I'm like, What does Linux mean?

Ah yeah good catch! I'm about 99% sure we can recreate the nesting there so it makes sense in this new context. I think I must've instructed Claude at some point to flatten the nesting so it didn't go too many levels deep, but it is warranted here.

sg-writer:
@joelhans alright I've restored most of the nesting where it was lost in places like the one you highlighted. It may still need some gentle massaging but now I think it's comprehensible at least.

S3Prototype (approved):
Gonna approve to unblock, with the understanding that the finer details will be addressed in the future 👍 

My overall notes:

1. I think this is significantly better than what we had before. It feels a lot clearer for the user to navigate this space and understand where to go after landing at the docs home.
2. I think in the sidebar, Network should go above Traffic. Domains and TCP Addresses are more fundamental things customers tend to want to know about
3. The "Secure and Control Access" subsection, I don't know why but reading that initially felt a little confusing to me. Likely because "Secure" is an adjective and a verb, and my brain didn't know how to instantly parse it. "Restrict and Control Traffic" feels a bit better IMO. Up for debate of course, but the fact that this is not where you'd find info about identity & access control definitely feels off since the name literally has "Control Access" in it

But overall, really great stuff

PR #2318, opened 2026-08-20
docs(agent): mark 3.20x through 3.22x as end of life

iambmelt:
The version support chart still listed 3.20x, 3.21x, and 3.22x as Active, but all three are past their End of Life date (2026-04-20, 2026-05-13, and 2026-05-20). This flips their Support Status to End of life to match the policy. 3.23x stays Active until its EoL on 2026-08-26.


iambmelt:
Note: this only updates the docs to reflect current support status — it does not lock these versions (3.20x–3.22x) out of the platform. They can still connect for now. I'll handle the actual enforcement/lockout in ~60 days; there's an active deprecation campaign in flight, so I'm holding off on cutting them off until that runs its course.

sg-writer (approved):

PR #2324, opened 2026-08-26
Fix inaccurate custom load-balancing docs for forward-internal

kate-osborn:
The endpoint_selectors and endpoint_weights fields referenced for custom balancing strategies don't exist in the forward-internal Traffic Policy action. Removed the false claims while keeping the Early Access callout for the upcoming feature.


Alice-Lilith (approved):

gateway/endpoints/endpoint-pooling.mdx
sg-writer:
```suggestion
## Protocol, binding, and type
```

PR #2328, opened 2026-08-27
Standardize the Share Localhost sidebar and reframe its how-it-works page

sg-writer:
Follow-up to #2316 to standardize the entry flow for all three products. Promotes the tunnels page to a flat entry page so all three products open Overview → Quickstart → How It Works, renames "The ngrok platform" group to "Concepts", and title-cases "Use Cases".

Also rewrites the tunnels page around the use case instead of the mechanism: it led with tunnel internals, and now leads with what you get. Retitling it to "How Sharing Localhost Works" also clears the name collision with `gateway/agent/overview` ("Secure Tunnels Overview" which probably needs a similar type of reframing).

### DIFF
--- a/share-localhost/tunnels.mdx
+++ b/share-localhost/tunnels.mdx
@@ -40,7 +40,7 @@ See the [protocols documentation](/gateway/endpoints/protocols) for the details 
 ## Who can reach it
 
 Your URL is public as soon as it exists.
-Anyone who has it can open it, so treat it like a link to a live service rather than a private address — and remember that whatever you share is running on your own machine.
+Anyone who has it can open it, so treat it like a link to a live service rather than a private address—and remember that whatever you share is running on your own machine.
 
 You control access without changing your app:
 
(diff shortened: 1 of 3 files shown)


share-localhost/tunnels.mdx
```diff
@@ -1,40 +1,58 @@
...
+| A web app, API, or webhook receiver | `ngrok http 8080` |
+| A database, SSH server, RDP host, or game server | `ngrok tcp 3389` |
+| A service that presents its own TLS certificate | `ngrok tls 8443` |
+
+See the [protocols documentation](/gateway/endpoints/protocols) for the details of each.
+
+## Who can reach it
+
+Your URL is public as soon as it exists.
+Anyone who has it can open it, so treat it like a link to a live service rather than a private address — and remember that whatever you share is running on your own machine.
```
sg-writer:
```suggestion
Anyone who has it can open it, so treat it like a link to a live service rather than a private address—and remember that whatever you share is running on your own machine.
```

S3Prototype (approved):

PR #2330, opened 2026-08-28
Reframe the agent overview around what it connects

sg-writer:
Reframes the agent overview around the use case, matching what #2328 did for the Share Localhost tunnels page. It opened with what the agent is — "a lightweight command-line program that creates secure tunnels" — and now opens with the problem it solves: the service you need to reach is behind a firewall, on a device in the field, or on a laptop, and can't take an inbound connection.

Adds a "When you need an agent" section, including when you don't and should use a Cloud Endpoint instead, which is the first question a reader actually has. Retitles from "Secure Tunnels Overview" to "Connect Services the Internet Can't Reach"; the sidebar label stays "Overview". The Features list became "What you get once it's connected", and the use-case cards are unchanged.

### DIFF
--- a/gateway/agent/overview.mdx
+++ b/gateway/agent/overview.mdx
@@ -87,7 +87,7 @@ Endpoints an agent creates are [Agent Endpoints](/gateway/endpoints/agent-endpoi
 		icon="globe"
 		href="/gateway/api-gateway/get-started/"
 	>
-		Create an API gateway with internal agent endpoints and centralized traffic management policies.
+		Create an API gateway with internal Agent Endpoints and centralized traffic management policies.
 	</Card>
 </Columns>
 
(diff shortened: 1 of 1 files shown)


S3Prototype (approved):
Great stuff!

gateway/agent/overview.mdx
sg-writer:
```suggestion
		Create an API gateway with internal Agent Endpoints and centralized traffic management policies.
```

gateway/agent/overview.mdx
```diff
@@ -1,52 +1,64 @@
 ---
-title: Secure Tunnels Overview
+title: Connect Services the Internet Can't Reach
 sidebarTitle: Overview
-description: Learn about ngrok's secure tunnel agent for exposing local services, connecting devices, and creating secure remote access.
+description: Learn when you need the ngrok agent to reach services behind firewalls, NAT, and private networks, and what it gives you once they are connected.
```
sg-writer:
```suggestion
description: Learn when to use the ngrok agent to reach services behind firewalls, NAT, and private networks—and what it gives you once they're connected.
```

PR #2335, opened 2026-08-31
Rewrite relative changelog links to absolute when generating the k8s snippet

sg-writer:
The k8s changelog snippet is copied verbatim from `ngrok-operator`'s `CHANGELOG.md`, which lives at that repo's root. Its relative links resolve on GitHub but 404 here, which is how the 0.24 upgrade guide link broke in `snippets/k8s/_changelog.mdx`.

The generator now rewrites relative link targets to absolute `ngrok-operator` URLs (`blob/main` for links, `raw.githubusercontent.com` for images), leaving absolute URLs, `mailto:`, protocol-relative, and anchor-only targets alone.

Two related fixes along the way. The workflow had its own inlined `curl | perl` copy of the fetch logic and never called `custom/scripts/fetch-k8s-changelog.js`, so the script had drifted into dead code — it now runs the script, so there is one implementation. And the script's `OUTPUT_PATH` was one directory short, resolving to `custom/snippets/k8s/_changelog.mdx`; it now writes to the real snippet path.

The regenerated snippet is included so the published link is fixed without waiting for the Monday cron. It is byte-identical to the previous file apart from the two link lines, confirming the node path reproduces the shell pipeline's output.

Upstream companion, making the links absolute at the source: ngrok/ngrok-operator#885.

### DIFF
--- a/.github/workflows/auto-gen-k8s-changelog.yml
+++ b/.github/workflows/auto-gen-k8s-changelog.yml
@@ -27,15 +27,12 @@ jobs:
           token: ${{ steps.generate-token.outputs.token }}
           ref: main
 
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: '24'
+
       - name: Fetch and Update Changelog
-        run: |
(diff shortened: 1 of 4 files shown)


S3Prototype (approved):
Nice

PR #2351, opened 2026-09-04
Document the native Gemini API surface on gateway.ngrok.ai

Megalonia:
Part of [AIG-1434](https://linear.app/ngrok/issue/AIG-1434/document-the-native-gemini-generatecontent-endpoint).


S3Prototype (approved):

nijikokun (approved):
Looks good but there are vale issues here as well

PR #2352, opened 2026-09-04
Sync the AI Gateway OpenAPI spec and document the new API surfaces

Megalonia:


openapi/ai-gateway.yaml
```diff
@@ -712,6 +736,93 @@ paths:
...
+        whitespace from the key, so `" k"` and `"k"` are one key.
+
+        A key longer than 255 characters is rejected with a 400 whose
+        `details.reason` is `invalid_idempotency_key`. A key sent with a
+        different `amount` is rejected with a 409 whose `details.reason` is
+        `idempotency_key_reuse`, and whose `details.purchase_id` names the attempt
+        the key already points at. A request that starts while another purchase for
+        the account is unresolved is rejected with a 409 whose `details.reason` is
+        `purchase_in_progress`. Poll the `details.purchase_id` it returns instead
+        of retrying.
```
nijikokun:
```suggestion 
        Start a charge using the account's saved card. A `200` response means the
        purchase has started, not that credits have been granted. Poll the GET
        endpoint until the purchase reaches a terminal state. Do not retry a
        `pending` purchase, since the charge may still be processing.

        The account must be provisioned for credit purchases and have a default
        payment method. If the account is not provisioned, use the checkout
        session endpoint instead. Cards cannot be added through this API and must
        first be saved in the ngrok dashboard.

        Send an `Idempotency-Key` with every purchase to prevent duplicate charges.
        Use a new key for each purchase attempt, including when retrying a declined
        purchase. Reusing a key returns the existing purchase instead of starting
        another charge.

        Idempotency keys:
        - Are limited to 255 characters.
        - Ignore leading and trailing whitespace.
        - Cannot be reused with a different `amount`. This returns
          `409 idempotency_key_reuse`.
        - Can be safely used across concurrent requests, which will all return the
          same purchase.

        If another purchase for the account is still unresolved, the request
        returns `409 purchase_in_progress`. Poll the returned
        `details.purchase_id` instead of retrying.

        Common errors include:
        - `400 metronome_not_provisioned`: Use the checkout session endpoint.
        - `400 no_default_payment_method`: Add a default payment method in the
          ngrok dashboard.
        - `400 invalid_idempotency_key`: The idempotency key is longer than 255
          characters.
        - `409 idempotency_key_reuse`: The key was already used with a different
          amount.
        - `409 purchase_in_progress`: Another purchase is still unresolved.
```

openapi/ai-gateway.yaml
```diff
@@ -712,6 +736,93 @@ paths:
...
+      summary: Get a credit purchase
+      description: |
+        Get a single credit purchase attempt by ID, so the caller can poll it to
+        a terminal state and read the decline reason recorded on failure.
+
+        A purchase leaves `pending` only when the payment gate resolves. An
+        asynchronous webhook records that, and a periodic sweep backstops a lost
+        webhook. Therefore `pending` means the outcome is not recorded yet. It does
+        not mean the charge is still in flight. A purchase can stay `pending` until
+        the next sweep runs, after the charge itself settles.
```
nijikokun:
```suggestion
        Get a credit purchase by ID. Use this endpoint to poll the purchase until
        it reaches a terminal state and to read the decline reason if it fails.

        `pending` means the payment outcome has not been recorded yet. A purchase
        may remain `pending` briefly after the charge has settled.
```

openapi/ai-gateway.yaml
```diff
@@ -916,6 +1064,27 @@ components:
...
+      in: header
+      required: false
+      description: |
+        Optional caller-supplied key that makes a retry safe. A later request
+        that repeats the key returns the purchase attempt the first request
+        started, instead of starting a second charge. The key is scoped to the
+        account and to this operation. ngrok keeps the key for as long as the
+        purchase attempt record exists, and those records are not deleted, so treat
+        a key as permanently used. Accepts up to 255 characters. An absent or empty
+        header disables replay protection.
```
nijikokun:
```suggestion
        Optional key that makes retries safe. Reusing the same key returns the
        existing purchase instead of starting a second charge.

        Keys are scoped to the account and this operation, can be up to 255
        characters, and should be treated as permanently used. If the header is
        missing or empty, replay protection is disabled.
```

nijikokun (commented):
Looks good, just some adjustments to the copy for readability and reducing some of the throat clearing.

nijikokun (approved):

nijikokun (approved):

PR #2353, opened 2026-09-04
Publish the request attempts endpoint and a guide to reading it

Megalonia:


ai-gateway/guides/why-a-request-failed-over.mdx
```diff
@@ -0,0 +1,160 @@
+---
+title: Find out why a request failed over
+sidebarTitle: Why a request failed over
+description: Trace one AI Gateway request through every upstream attempt to see which providers it tried, in what order, and which one served it.
+---
+
+When the gateway answers, it does not say how much work that took.
+A response can come from the first upstream attempt or the fourth.
+This guide walks from a single response header to the full attempt chain behind it.
```
nijikokun:
```suggestion
The AI Gateway may make several attempts before returning a response. The response alone doesn’t show how many attempts were made. This guide shows you how to use a response header to find each attempt and see what happened.
```

ai-gateway/guides/why-a-request-failed-over.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+## 1. Read the attempt count off the response
+
+Every gateway response carries `Ngrok-AIG-Attempts`.
+A value above 1 means the first candidate failed and the gateway moved on.
```
nijikokun:
```suggestion
Every response through ngrok.ai includes the `Ngrok-AIG-Attempts` header, which shows how many attempts were made. A value greater than `1` means the first attempt failed and the gateway tried again.
```

ai-gateway/guides/why-a-request-failed-over.mdx
```diff
@@ -0,0 +1,160 @@
...
+ngrok-aig-model: gemini-3.6-flash
+```
+
+Three attempts, and the third one served.
+The served model is the third entry in the candidate list, so both earlier candidates failed.
+
+<Note>
+Read the model from `Ngrok-AIG-Model`, not from the `model` field in the response body.
+The gateway forwards bodies byte-for-byte, so the body keeps whatever the provider wrote there.
+On a multi-model failover the two can disagree, and the header is the authoritative one.
```
nijikokun:
```suggestion
Use the `Ngrok-AIG-Model` response header to identify the model that handled your request. The AI Gateway passes the provider’s response body through unchanged, including its `model` field. If the gateway switches models after a failed attempt, these values may differ. Use the header as the source of truth.
```

ai-gateway/guides/why-a-request-failed-over.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+```http
+HTTP/2 200
+ngrok-aig-request-id: rq_2abcDEF456
+ngrok-aig-attempts: 3
+ngrok-aig-model: gemini-3.6-flash
+```
+
+Three attempts, and the third one served.
+The served model is the third entry in the candidate list, so both earlier candidates failed.
```
nijikokun:
```suggestion
The AI Gateway made three attempts. The third candidate in the list handled the request after the first two candidates failed.
```

ai-gateway/guides/why-a-request-failed-over.mdx
```diff
@@ -0,0 +1,160 @@
...
+      "retryAction": "succeeded",
+      "completion": "succeeded",
+      "upstreamDurationMs": 1902
+    }
+  ]
+}
+```
+
+The endpoint is read-only and scoped to your account.
+A request belonging to another account reads as missing rather than forbidden.
```
nijikokun:
```suggestion
This endpoint lets you only view requests from your account. It cannot modify them. If a request belongs to another account, the endpoint returns “not found” instead of “forbidden.”
```

ai-gateway/guides/why-a-request-failed-over.mdx
```diff
@@ -0,0 +1,160 @@
...
+| `next_backend` | The provider or model failed, so the gateway moved to the next model in the failover list |
+| `managed_fallback` | A customer-owned key failed, so the gateway fell back to an ngrok-managed key |
+| `retry_same_backend` | The provider asked for a retry with `Retry-After`, so the gateway waited and retried the same backend |
+| `fail_fast` | The failure was not retryable, so the gateway stopped |
+| `exhausted` | No candidate remained |
+
+`retryAction` is optional.
+It is absent on attempts the gateway recorded before it began storing the verdict.
+
+In the example above, `next_key` on attempt 1 followed by `next_backend` on attempt 2 reads as: the OpenAI key hit a rate limit and no other OpenAI key was left to try, so the gateway advanced to the next model, whose key was rejected outright.
```
nijikokun:
```suggestion
## 3. Check why the AI Gateway retried or stopped

For each attempt, check `completion` to see how it ended and `retryAction` to see what the AI Gateway decided to do next. Use `served` to identify the attempt that returned the response to your client.

### Identify the attempt that returned the response

Look for `served: true`. At most one attempt per request has this value.

Do not assume the last attempt returned a response. A request can end with a failed attempt that did not produce a response for your client.

### Check how each attempt ended

The `completion` field has one of these values:

- `succeeded`
- `failed`
- `timed_out`
- `stream_failed`
- `unknown`

For failed attempts, read the error details to find the cause, then check `retryAction` to understand why the gateway retried or stopped.

### Follow the retry decisions

| `retryAction` | What happened |
| --- | --- |
| `succeeded` | The attempt produced the response your client received. |
| `next_key` | The credential failed, so the gateway tried the next key for the same provider. |
| `next_quota_pool` | The credential’s quota pool was exhausted, so the gateway moved to another pool. |
| `next_backend` | The provider or model failed, so the gateway tried the next model in the failover list. |
| `managed_fallback` | A customer-owned key failed, so the gateway switched to an ngrok-managed key. |
| `retry_same_backend` | The provider requested a retry using `Retry-After`, so the gateway waited and retried the same backend. |
| `fail_fast` | The failure could not be retried, so the gateway stopped. |
| `exhausted` | No candidates remained to try. |

`retryAction` is optional. It is absent from attempts recorded before the gateway began storing retry decisions.

In the example above, `next_key` on attempt 1 means the gateway tried another key for the same provider. `next_backend` on attempt 2 means it then moved to the next model in the failover list. Check each attempt’s error details to see why it failed.
```

ai-gateway/guides/why-a-request-failed-over.mdx
```diff
@@ -0,0 +1,160 @@
...
+| `next_key` on every OpenAI attempt | Add more [provider keys](/ai-gateway/guides/attaching-provider-keys) for that provider, or raise your quota with them |
+| `next_backend` on the first candidate every time | Your primary model is the wrong default. Reorder the [fallback list](/ai-gateway/guides/configure-fallback-models) |
+| A single `fail_fast` attempt | The request itself was rejected. Check the status code and the [error code](/ai-gateway/reference/error-codes); failover would not have helped |
+| `exhausted` at the end | Every candidate failed. Widen the candidate list or attach credentials for more providers |
+| `managed_fallback` you did not expect | Your own key is failing and ngrok credits are absorbing the traffic. Check the key |
+
+## What is not on this endpoint
+
+Captured request and response bodies.
+Those hold the prompt and the answer, so they stay in [app.ngrok.ai](https://app.ngrok.ai) rather than on the API.
```
nijikokun:
```suggestion
## What this endpoint does not include

This endpoint does not include captured request and response bodies, which could contain sensitive information.
```

nijikokun (approved):
Requested some changes, approved after those are in

PR #2359, opened 2026-09-09
Subgroup the Gateway Agents sidebar and wire its reference pages

sg-writer:
Splits the flat Agents list into Guides, Configuration File, Reference and Versions & Upgrades, and wires the agent's CLI, API, config file and changelog into the section instead of only the Reference tab. The three Docker pages move to Integrations › General Guides with the other `using-ngrok-with` guides. No page moves on disk and no new redirects.

The group's `root` is set to `gateway/agent/index` so the Overview can lead the sidebar without `/gateway/agent` redirecting off the page that owns its 41 anchors. Also relabels the entry that said "Quickstart" but pointed at TLS termination, and repairs two dead anchors into `/gateway/agent/`.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -499,8 +499,8 @@
               {
                 "group": "Agents",
                 "pages": [
-                  "gateway/agent/index",
                   "gateway/agent/overview",
+                  "gateway/agent/index",
                   {
                     "group": "Guides",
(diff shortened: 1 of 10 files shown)


S3Prototype (approved):
LGTM

